PR #17, opened 2019-01-15
[CI] Update tests to run on PR's

shepting:
## Changes
- fixes `make debug` (update device name to match Xcode 10)
- limit threads to 4 to see build progress more easily


.buckconfig
```diff
@@ -5,10 +5,9 @@
   combined_preprocess_and_compile = true
   pch_enabled = false
   ldflags = -Xlinker -objc_abi_version -Xlinker 2 -fobjc-arc -fobjc-link-runtime
-  
 
 [swift]
-  version = 4.2
```
dfed: 😢, but makes sense

App/BUCK
```diff
@@ -16,7 +16,10 @@ apple_resource(
 
 apple_binary(
     name = "ExampleAppBinary",
-    visibility = ["//App:"],
+    visibility = [
+        "//App:",
+        "//App/...",
```
dfed: what does this line accomplish?

App/BUCK
```diff
@@ -27,11 +30,9 @@ apple_binary(
         "//App/Tests:Tests",
     ],
     deps = [
-        "//Libraries/Swift2:Swift2",
```
dfed: Why are we removing `Swift2`?

App/BUCK
```diff
@@ -27,11 +30,9 @@ apple_binary(
...
         "//Libraries/Swift3:Swift3",
         "//Libraries/Objc1:Objc1",
         "//Pods:CryptoSwift",
-        "//Pods:OHHTTPStubs",
```
dfed: Why are we removing `OHHTTPStubs`?

Makefile
```diff
@@ -18,11 +19,14 @@ build:
 	$(BUCK) build //App:ExampleAppBundle
 
 debug:
-	$(BUCK) install //App:ExampleAppBundle --run
+	$(BUCK) install //App:ExampleAppBundle --run --simulator-name 'Phone: iPhone Xs'
```
dfed: is the simulator name not cap-sensitive? I thought this should be `iPhone XS`.

App/BUCK
dfed: Note that in our app's BUCK file all we have is `visibility = ["PUBLIC"]`. Should we change that?

.buckconfig
bachand: Why does it make sense?

.travis.yml
```diff
@@ -9,8 +9,4 @@ before_install:
 env:
   - TERM=dumb
 script: 
-  - make targets
-  - make build
-  - make test
-  - make project
-  - make xcode_tests
+  - make ci
```
bachand: 👍

App/BUCK
bachand: I think this is saying that this is now visible to any BUCK target under `//App`.

In our main app I think we will want to move away from `visibility = ["PUBLIC"]` as we have started to do in our platform work.

.buckconfig
bachand: Ah got it. This aligns with our version of Xcode I guess.

Makefile
bachand: https://github.com/airbnb/BuckSample/pull/17/commits/04df820e7aeebd9ccad7552c4b214501f4c561c9

PR #20, opened 2019-01-29
Update Buck version

dfed:
Points us at the buck version from https://github.com/facebook/buck/pull/2163

Also makes the `Makefile` use this version of `buck` for consistency.


bachand:
👍

PR #27, opened 2019-01-31
Introduce suppress_warnings flag

dfed:
`warning_as_error` is a bit too binary. This new `suppress_warnings` allows us to opt individual modules into allowing warnings without suppressing them.


bachand:
Glad we brought this over

PR #28, opened 2019-01-31
Add a module that depends on a library with Objective-C++ in it

dfed:


bachand:
Didn’t review this in detail but this is a good change. The more real we can make this project the better our lives will be IMHO

PR #29, opened 2019-02-01
Add library with an asset catalog

dfed:
Adds a library with an asset catalog, where a resource from that asset catalog is referenced from unit tests.

~Running unit tests for this new library in Xcode fails. It seems to work on the command line, however.~

~If anyone knows how to get this to work, help would be greatly appreciated.~

This library's unit test target works both from command line and generated Xcode project


dfed:
Looks like to get this working in Xcode, the tests need a hosted test app, and the test target needs a bundle identifier. Trying to figure out how to get a bundle identifier picked up

dfed:
Looks like the hosted test app wasn't required. What was required was pumping the binary config into the unit test target.

bachand:
@dfed it looks like CI is failing?

App/ViewController.swift
```diff
@@ -57,6 +58,9 @@ class ViewController: UIViewController {
     // Swift class imports a library with Objective-C++
     MySwiftReliesOnCXXClass().doSomethingWithCXXLibrary()
 
+    // This line will crash if the assets haven't been bundled into the app
```
bachand: Let's specify what library this is coming from, just for good measure.

Libraries/SwiftWithAssets/Tests/SwiftWithAssetsTests.swift
```diff
@@ -0,0 +1,12 @@
+import XCTest
+
+@testable import SwiftWithAssets
+
+final class SomeSwift3Tests: XCTestCase {
+
+  func test_buckImageExists() {
+    _ = Catalog.buck.image
```
bachand: Just for good measure, can we add some `XCTAssert...` function in here?

I realize it doesn't really matter, but I think that would make things more clear and self-evident for someone reading this.

Libraries/SwiftWithAssets/Tests/Info.plist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bachand: I guess tests need their own `Info.plist` since they're literally a separate executable?

dfed: That's my understanding.

Libraries/SwiftWithAssets/Tests/SwiftWithAssetsTests.swift
dfed: Yeah I'll come up with something. Since `image` is not optional it's kinda hard to write an XCTAssert that uses it. But I can put something in there to make it clear that tests are succeeding.

Issue #2, opened 2017-10-31
.ipa path


hakanyildizay:
Hi everybody,
I am trying to export ipa file.

Here is .buckconfig
`[cxx]
  default_platform = iphonesimulator-x86_64
  cflags = -g -fmodules -fobjc-arc -D BUCK -w
  combined_preprocess_and_compile = true

[swift]
  version = 3.2
  compiler_flags = -DBUCK -whole-module-optimization -enable-testing -suppress-warnings
`

And BUCK file

`apple_resource(
    name = 'AppResources',
    dirs = [],
    files = glob(['**/*.xib','**/*.storyboard']),
)

apple_asset_catalog(
  name = 'AppAsset',
  dirs = ['SampleApp/Assets.xcassets'],
)

apple_binary(
    name = 'AppBinary',
    srcs = glob([
        'SampleApp/*.swift',
      ]),
    deps = [':AppResources', ':AppAsset'],
    frameworks = [
        '$SDKROOT/System/Library/Frameworks/Foundation.framework',
        '$SDKROOT/System/Library/Frameworks/UIKit.framework',
    ],
)

apple_bundle(
    name = 'TheApp',
    binary = ':AppBinary',
    extension = 'app',
    info_plist = 'SampleApp/Info.plist',
    info_plist_substitutions = {
        'PRODUCT_BUNDLE_IDENTIFIER': 'com.some.bundleID',
        'CURRENT_PROJECT_VERSION': '1',
        'DEVELOPMENT_LANGUAGE': 'Base',
    },
)

apple_package(
  name = 'SampleAppPackage',
  bundle = ':TheApp',
)
`

I don't know what to pass to apple_package() or apple_bundle() functions. 
In configuration file I couldn't find .ipa export or archive path.
  
I would really appreciate if you could help me.
Thank you.

zayhero:
This config looks fine, if you run `buck build //:SampleAppPackage`, you should able to find the `.ipa` file in `buck-out/gen/SampleAppPackage`

PR #31, opened 2019-02-02
Simpler library declaration, and faster cI

dfed:
This is a pretty large change. We don't need to land this all at once if we don't want to.

This PR accomplishes two things.

1. Greatly simplifies how we declare a `library`. I added a `library` macro that should be used for libraries defined by this project (not 3rd-party libraries). This macro automatically creates a unit test target for the library, handles finding sources (it assumes that sources are in `Sources/` and test sources in `Tests/`). Having a single `library` macro that creates both an `apple_library` and an `apple_test` target allows libraries to need minimal configuration, and also encourages the creation of unit tests per new library.
2. Speeds up CI tests. Since creating `apple_test` targets can be slow in CI, the `library` macro creates an additional `apple_library` for every unit test in the app. I then created a `apple_test_all` macro that can create a single `apple_test` for all the test targets in the app. Applying this strategy of creating only a single `apple_test` in CI saved more than 30 minutes of build time per PR at Airbnb. Shout out to @zayhero who was the first Airbnb engineer to come up with the concept of using a single `apple_test` in CI.


bachand:
😃

App/BUCK
```diff
@@ -26,20 +39,22 @@ apple_binary(
...
-    ],
+    ] + libraryDependencies,
+)
+
+# This test to bundles all unit test libraries into a single test target.
+# Test targets can be slow to create in CI; creating only one can save significant time.
+apple_test_all(
+    name = "ExampleAppCITests",
+    info_plist = "Info.plist",
+    srcs = ["Dummy.swift"],
```
bachand: What if this is an implementation detail of `apple_test_all`, so that the call looks like:
```python
# This test to bundles all unit test libraries into a single test target.
# Test targets can be slow to create in CI; creating only one can save significant time.
apple_test_all(
    name = "ExampleAppCITests",
    info_plist = "Info.plist",
    srcs = ["Dummy.swift"],
    libraries = libraryDependencies,
    additional_tests = appTests,
)
```

I feel like this will be something easy to forget to add.

bachand: If we did that I wonder if we should also check that the file actually exists to avoid some cryptic build errors? 🤷‍♂️

bachand: Ideally I guess Buck would fail with a reasonable error message.

dfed: I like the idea of abstracting `srcs` away, and also hiding the Dummy file in some kind of Support folder that isn't at the root.

dfed: So, defining the `srcs` in our `bzl` file doesn't work, since `srcs` will always be relative to the calling function.

But, there's good news! Looks like the buck test target doesn't actually need any `srcs` to run. I think we need `srcs` for Xcode, but not necessarily for Buck.

Config/buck_rule_macros.bzl
```diff
@@ -8,23 +8,71 @@ def apple_third_party_lib(**kwargs):
...
+        labels = [],
+        **kwargs):
     test_name = name + ".test"
+
+    if bundle_for_ci:
+        # Create a library with the test files. We'll use use these for our CI tests.
+        # Libraries are much faster to create in CI than unit test bundles.
+        # Therefore, we package up this test target as a library that we can depend on
+        # later with a single apple_test bundle.
+        native.apple_library(
```
bachand: Why do we need `native.` here?

Config/buck_rule_macros.bzl
```diff
@@ -62,6 +110,84 @@ def apple_lib(
...
+        has_objective_c = False,
+        internal_headers = None,
+        extra_xcode_files = [],
+        deps = [],
+        frameworks = [],
+        info_plist = "Tests/Info.plist",
+        info_plist_substitutions = {},
+        test_host_app = None,
+        run_test_separately = False,
+        additional_tests = [],
```
bachand: When do we think this would be used? Not saying it would or wouldn’t be. I just don’t see any current usage and I want to understand what you had in mind.

Libraries/Objc1/Tests/Info.plist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bachand: Is it your understanding that in our end state we will want each test target to have it’s own `Info.plist` file? Part of me feels like it’s something that could be shared but I am not sure a) if that’s feasible, or b) if there are reasons why we wouldn’t want to go that route.

Config/buck_rule_macros.bzl
dfed: I don't know. Looks like it's referencing a library or object called `native`? The project doesn't work without it.

Config/buck_rule_macros.bzl
dfed: It's solely here for flexibility. If someone defined tests outside of a `Tests` directory, they could do so here.

It was probably overkill to include it. Let's include it when we have a use case.

Libraries/Objc1/Tests/Info.plist
dfed: I'd love to share. I haven't yet figured out how. There are two big blockers to sharing a single info.plist file:
1. All `info_plist` paths are relative to the calling BUCK file
2. Buck complains if you try to include a file that is closer in the file system to another BUCK file than the one you are referencing that file from.

To better explain #2, imagine your repository has the following files on the file system:
```
//BUCK
//BuckSupport/SharedInfo.plist
//Libraries/BUCK
```
Now, say in `//Libraries/BUCK` you try to reference `../BuckSupport/SharedInfo.plist`. Buck will fail, telling you that `SharedInfo.plist` should be managed by `//BUCK`, not `//Libraries/BUCK`, since `//BUCK` is closer to that file.

One idea (that I haven't tested) is to pass in a dictionary rather than a file for an Info.plist – as far as I can tell there's no great reason to have an actual file on disk. If using a dictionary is possible, that'd solve our problem. If it isn't, it feels like something we could contribute back to FB's `buck`.

Config/buck_rule_macros.bzl
```diff
@@ -8,23 +8,71 @@ def apple_third_party_lib(**kwargs):
         **kwargs
     )
 
-def apple_test_lib(name, **kwargs):
+def test_name(name):
+     return name + "Tests"
+def ci_test_name(name):
+    return name + "-For-CI"
+
+def apple_test_lib(
```
bachand: I liked the convention we were using of prefixing our macros with `airbnb`. Do you think we should do that here?

Config/buck_rule_macros.bzl
bachand: Weird. I wonder if it needs to disambiguate for some reason.

Config/buck_rule_macros.bzl
bachand: 👍

Config/buck_rule_macros.bzl
dfed: I was avoiding using `airbnb`-specific references in this repo. I do think some kind of prefix to denote that this is a special macro would be useful. We just need to come up with one (and `airbnb` may be a reasonable one. Though maybe this is where Bob the Builder comes in handy). Let's chat offline and come up with a good prefix, then do another pass 🙂

Config/buck_rule_macros.bzl
```diff
@@ -1,4 +1,4 @@
-load("//Config:configs.bzl", "test_configs", "library_configs", "pod_library_configs")
+load("//Config:configs.bzl", "test_configs", "library_configs", "pod_library_configs", "binary_configs")
```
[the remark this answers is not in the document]
dfed: Will do before merging.

Libraries/Objc1/Tests/Info.plist
bachand: All makes sense. Thanks for the context.

Config/buck_rule_macros.bzl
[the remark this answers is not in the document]
bachand: Something we’ll have to do in `apps` too down the road I guess

App/BUCK
dfed: The `apple_test_all` is entirely ignored by Xcode when developing the main app. 

We could theoretically generate a workspace for this target by running `buck project //App:ExampleAppCITests`, but I don't know why we would. One of the benefits to how we've set this up (with the example app not depending at all on these CI-Test jobs) is that this target lives outside the world we want to generate with Xcode.

Config/buck_rule_macros.bzl
```diff
@@ -62,6 +110,84 @@ def apple_lib(
...
+        lib_test_name,
+        srcs = test_sources,
+        headers = test_headers,
+        info_plist = info_plist,
+        info_plist_substitutions = info_plist_substitutions,
+        test_host_app = test_host_app,
+        run_test_separately = run_test_separately,
+        frameworks = test_frameworks,
+        deps = [":" + name] + test_deps,
+        **kwargs)
```
bachand: With this change is there any way to run the tests for just a single module, e.g. `Swift2`? Not saying there necessarily has to be, but I wanted to understand.

Config/buck_rule_macros.bzl
dfed: - [ ] Remove additional_tests

App/BUCK
dfed: And to be a bit more explicit, we are still capable of running our unit tests in Xcode. What I've done with `apple_test_all` is create a distinct way of running unit tests, one that is built to work with Buck (caring not at all about Xcode compatibility), that is meant to be used only in CI.

bachand: I guess I’m maybe confused: is it possible to run these tests in Xcode? If so, how is that being done?

Config/buck_rule_macros.bzl
bachand: Nvm, digging into this more it looks like we’re creating a way to do that here.

App/BUCK
bachand: OK so here’s my understanding, upon looking closer:

- `first_party_library` creates a test target for each library. That must be what’s used when I want to test a module (or all modules) in the generated Xcode app. I’m assuming I could just do <kbd>CMD</kbd> + <kbd>U</kbd> on the main app scheme and it would run _all_ the tests, or I could open a scheme for just one library and hit <kbd>CMD</kbd> + <kbd>U</kbd> to run tests for just that library.
- We have `apple_test_all` to create one giant test target for running all of the test in CI that avoids the overhead of creating a test target per library.

Is that about right? Just trying to make sure I’m 100% on the same page.

dfed: I added pretty extensive comments to the main macros we're using. PTAL!

Libraries/Objc1/Tests/Info.plist
dfed: Looks like `Info.plist` _must_ be a file. Let's add this to our tasks of "things that we should contribute back to Buck"

PR #34, opened 2019-02-05
Specify SWIFT_VERSION as 4.0 instead of 4

dfed:
When `SWIFT_VERSION` is set to `4` rather than `4.0`, `mlmodel`'s fail to compile.


bachand:
Good find

PR #35, opened 2019-02-06
Import Swift from Objective-C

dfed:
Proving this is possible in a library.


bachand:
Good idea.

We can save the debate over `[[Foo alloc] init]` v `[Foo new]` for another day :wink:

PR #37, opened 2019-02-06
Check warnings flags once, not twice

dfed:
Don't repeat yourself

cc @shepting


bachand:
Can you explain why this is no longer necessary? Can you include a link to the other place where we do this? I’m presuming that’s why we’re deleting.

bachand:
Thanks for adding that comment. I just want these PRs to be self-documenting.

PR #38, opened 2019-02-06
Add example of https://bugs.swift.org/browse/SR-6004, and instructions on how to work around it

bachand:
Good find @erichoracek!

cc: @dfed @fdiaz


bachand:
This is ready for review

App/ViewController.swift
```diff
@@ -61,6 +62,32 @@ class ViewController: UIViewController {
...
+    //
+    // This is tracked by https://bugs.swift.org/browse/SR-6004. This seems to indicate that
+    // `-all_load` will work too, but we haven't verified that.
+//    let myObject = MyPublicClass()
+    let myObject = MyFactory.myPublicObject()
+    if (myObject as? MyPublicProtocol) == nil {
+      print("Incorrect: `MyPublicProtocol` conformance is being erroneously stripped")
+    } else {
+      print("Correct: `MyPublicProtocol` conformance is not being stripped")
+    }
```
dfed: is it worth writing the above as a unit test?

bachand: I think that's a good idea. I'm gonna leave as-is for now but something to think about down the road. 🤞 Apple fixes this soon since it's affecting both Airbnb and Lyft (and this presumably other big companies as well!)

PR #44, opened 2019-03-04
 Make application tests @testable import ExampleApp

### LINKED ISSUE #24: @testable import ExampleAppBinary doesn't work
I've mentioned this issue on Slack channel, and wanted to try it with AirBnB's custom fork of `buck`.

Still no luck though.

Adding `@testable import ExampleAppBinary` to any of the tests, e.g. to `BuckSampleTests.swift` or `BuckSampleUITests.swift` causes this error:

```
Command failed with exit code 1.
stderr: App/UITests/BuckSampleUITests.swift:10:18: error: no such module 'ExampleAppBinary'
@testable import ExampleAppBinary
                 ^
```

While I can definitely find the `ExampleAppBinary` module with `find buck-out -name "*ExampleApp*"`

```
buck-out/gen/App/ExampleAppBinary#iphonesimulator-x86_64,swift-compile
buck-out/gen/App/ExampleAppBinary#iphonesimulator-x86_64,swift-compile/ExampleAppBinary.swiftmodule
buck-out/gen/App/ExampleAppBinary#iphonesimulator-x86_64,swift-compile/ExampleAppBinary-Swift.h
buck-out/gen/App/ExampleAppBinary#iphonesimulator-x86_64,swift-compile/ExampleAppBinary.o
```

I can't use it in Unit or UI tests though.

The original issue is here: https://github.com/facebook/buck/issues/2074

P.S.
I know it's open source and all that, but I don't have the time and code base knowledge to locate the change that removed this functionality and revert just the bits that make app module imports work again.

Since this is the only up-to-date buck demo of mix-n-match project there is (🥇), I thought raising it here would draw some attention to it.

Also, I though most of AirBnB's changes made it to master already... At least I was able to use `import` statements in our mix-n-match project with `modular=True` flag using current `master`.

dfed:
Resolves #24

### DIFF
--- a/App/Tests/BuckSampleTests.swift
+++ b/App/Tests/BuckSampleTests.swift
@@ -6,13 +6,14 @@
 //
 
 import XCTest
-@testable import Swift2
+@testable import ExampleApp
 
 final class BuckSampleTests: XCTestCase {
 
-    func testMySwiftClassAdding() {
-        let sut = MySwiftClass()
-        XCTAssertEqual(sut.add(1, 2), 3)
+    func testViewController() {
+        // This test accesses internal classes and properties in the ExampleApp library.
+        let sut = ViewController()
+        XCTAssertEqual(sut.value, "ViewName")
     }
 
     func testExample() {
--- a/App/BUCK
+++ b/App/BUCK
@@ -29,13 +29,13 @@ first_party_library_dependencies = [
     "//Libraries/SwiftWithMLModel:SwiftWithMLModel",
 ]
 
-apple_binary(
-    name = "ExampleAppBinary",
+apple_library(
+    name = "ExampleApp",
     visibility = [
         "//App:",
         "//App/...",
     ],
-    configs = binary_configs("ExampleApp"),
+    configs = library_configs(),
     swift_version = "4.0",
     srcs = [
         "ViewController.swift",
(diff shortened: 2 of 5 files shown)


App/Tests/BuckSampleTests.swift
```diff
@@ -6,13 +6,13 @@
 //
 
 import XCTest
-@testable import Swift2
+@testable import ExampleApp
 
 final class BuckSampleTests: XCTestCase {
 
-    func testMySwiftClassAdding() {
```
dfed: This test was moved into `SomeSwift2Tests`

bachand:
Can we verify that the debugger still works with this change? I know that @shepting made some headway on fixing the debugger by moving code from the library into the binary, so I want to be sure we didn't regress that work. Testing in the VC should be fine.

dfed:
Given that our app is currently using this structure I didn't test the debugger explicitly before merging. But I've now validated that the debugger continues work with this change. I think the key to the debugger working was making the binary include a swift file rather than an objc file.

bachand:
Thanks for checking 🙏 

Makes some sense re: source file👌

bachand:
(As much as this stuff ever makes sense haha 🙃)

dfed:
When we moved to having the app code in the main binary, we were effectively introducing swift into the main binary, which likely had the same effect as having a single dummy Swift source file in the binary. I think. Def a bit of conjecture here 😬

bachand:
I’ll buy it 😉

Issue #43, opened 2019-02-28
`apple_test_all` won't run when pre-compiled libraries are included


dharr419-ford:
This is a bit of a non-intuitive thing about Buck in general where pre-compiled libraries aren't auto-embedded, so it might be worth including as an example. Makes sense since you have to add the Copy Files phase in Xcode, but it _feels_ like Buck would do that for you but doesn't (there are issues logged around this in the Buck repo).

Biggest consequence of that is you have to redefine those dependencies to get `apple_test_all` to work and not crash at runtime, which it doesn't have the flexibility to do.

Thinking redefining it as:
```
def apple_test_all(
        libraries = [],
        additional_tests = [],
        additional_deps = [],
        **kwargs):
    ci_test_libraries = []
    for library in libraries:
        ci_test_libraries.append(ci_test_name(test_name(library)))

    apple_test_lib(
        deps = ci_test_libraries + additional_tests + additional_deps,
        bundle_for_ci = False,
        **kwargs
    )
```

should work fine.

dfed:
I think this could be solved by no longer using `libraries` and instead declare your pre-built libraries directly as `deps`. I’m not positive though. Can confirm next week, but I think we moved from `libraries` recently due to a conversation in the Buck Slack.

dfed:
Ah, I think I misread the OP here. @dharr419-ford looks like you're saying that you are already declaring your precompiled libraries as `deps`, but they aren't being embedded transitively into your CI test target. Is your precompiled a library a static library or a dynamic library? I haven't seen this behavior before, so I'm trying to get a better sense of the issue.

dfed:
In https://github.com/airbnb/BuckSample/pull/46, it didn't seem like I needed to do anything special in `apple_test_all` to support the `SwiftWithPrecompiledDependency` library. I'm curious how your configuration differs!

thedavidharris:
Dynamic, currently adding Buck onto a project that we've split out to a number of modules prebuilt with Carthage. Should be able to take a look at this in the next day or two.

thedavidharris:
So here's a simple way to reproduce the usual `Library not loaded` crash:

Cartfile:
`github "krzyzanowskim/CryptoSwift"`

Carthage/Buck:
```
prebuilt_apple_framework(
    name = "CryptoSwift",
    framework = "Build/iOS/CryptoSwift.framework",
    preferred_linkage = "shared",
    visibility = ["PUBLIC"],
)
```

[...]

PR #46, opened 2019-03-05
Add a library with a precompiled dependency

dfed:
We didn't have one, and now we do.


bachand:
We are holding off on merging this PR until we resolve the issues outlined in https://github.com/airbnb/BuckSample/issues/43

PR #52, opened 2019-03-15
Add  dynamic framework to BuckSample

qyang-nj:
This PR unlocks the ability of adding dynamic frameworks which can be shared between bundles (App, Extension or Framework). It works in Buck CLI and generated workspace.

A shared module named `SwiftShared` is added and imported into `ExampleApp` and `ExampleMessageExtension`. 

The bare bone of the BUCK rules look like this:

- The BUCK file for `SwiftShared`:
``` python
apple_library(
    name = "SwiftShared",
    # Setting preferred_linkage to shared makes this a dylib. 
    preferred_linkage = "shared",
    # Set the install_name so consumers of this dylib know where to find it.
    linker_flags = ["-Wl,-install_name,@rpath/%s.framework/%s" % (name, name)],
    configs = { "Debug": {
        # Make this a dylib in Xcode
        "MACH_O_TYPE": "mh_dylib",
    }},
)

apple_bundle(
    name = "SwiftSharedFramework",
    # Specifying the the flavor "shared" makes this a dynamic framework
    binary = ":SwiftShared#shared",
    extension = "framework",
)
```
- The BUCK file for the app:
``` python
apple_binary(
    name = "ExampleAppBinary",
    # The app binary depends on the library
    deps = [
         ":SwiftShared",
    ],
)

apple_bundle(
    name = "ExampleApp",
    binary = ":ExampleAppBinary",
    extension = "app",
    # the app bundle depends on the framework
    deps = [":SwiftSharedFramework"],
)
```
A library is a file where binary code lives, while a framework is a directory (bundle) containing a library and some resources. `ExampleAppBinary` depends on `SwiftShared` library because the binary needs to know where the actual code is, and the dependency from `ExmapleApp` bundle to `SwiftSharedFramework` simply copies the framework to app bundle (`ExmapleApp.app`). That being said, an extension only needs to depend on the library to compile, not the framework, since the main app bundle is the one who copies over the framework.

## Verification
* binary from Buck CLI build has undefined symbols from `SwiftShared`
image

* binary from Xcode build has undefined symbols from `SwiftShared`
image

### DIFF
--- a/App/BUCK
+++ b/App/BUCK
@@ -75,9 +75,6 @@ apple_library(
     ]
     + first_party_library_dependencies
     + build_phase_scripts,
-    # The library doesn't necessarily need to depend on frameworks for Buck CLI.
-    # It's a hack to make generated project link to the frameworks.
-    + first_party_frameworks,
 )
 
 apple_binary(
--- a/App/BUCK
+++ b/App/BUCK
@@ -1,6 +1,5 @@
 load("//Config:configs.bzl", "app_binary_configs", "library_configs", "watch_binary_configs", "message_binary_configs", "pretty", "info_plist_substitutions", "bundle_identifier", "DEVELOPMENT_LANGUAGE")
 load("//Config:buck_rule_macros.bzl", "apple_lib", "apple_test_lib", "apple_test_all")
-load("//Config:frameworks.bzl", "first_party_frameworks")
 
 apple_asset_catalog(
     name = "ExampleAppAssets",
@@ -75,9 +74,6 @@ apple_library(
     ]
     + first_party_library_dependencies
     + build_phase_scripts
-    # The library doesn't necessarily need to depend on frameworks for Buck CLI.
-    # It's a hack to make generated project link to the frameworks.
-    + first_party_frameworks,
 )
 
 apple_binary(
@@ -121,8 +117,11 @@ xcode_postbuild_script(
 apple_test_all(
     name = "ExampleAppCITests",
     libraries = first_party_library_dependencies,
-    additional_tests = app_tests + first_party_frameworks,
+    additional_tests = app_tests,
     prebuilt_frameworks = prebuilt_frameworks,
+    deps = [
+        "//Libraries/SwiftShared:SwiftSharedFramework",
+    ],
 )
 
 xcode_workspace_config(
@@ -150,7 +149,9 @@ apple_bundle(
     product_name = "ExampleApp",
     info_plist = "Info.plist",
     info_plist_substitutions = info_plist_substitutions("ExampleApp"),
-    deps = first_party_frameworks + [
+    deps = [
+        "//Libraries/SwiftShared:SwiftSharedFramework",
+    ] + [
         # For "#watch", https://buckbuild.com/rule/apple_bundle.html#deps
         ":ExampleWatchApp#watch",
         ":ExampleMessageExtension",
--- a/App/BUCK
+++ b/App/BUCK
@@ -312,6 +312,7 @@ apple_test_lib(
     ]),
     deps = [
         ":ExampleAppLibrary",
+        "//Libraries/SwiftShared:SwiftSharedFramework",
     ]
     + prebuilt_frameworks,
 )
--- a/App/BUCK
+++ b/App/BUCK
@@ -74,7 +74,10 @@ apple_library(
         "//App/Resources:StoryboardResources",
     ]
     + first_party_library_dependencies
-    + build_phase_scripts,
+    + build_phase_scripts
+    # The library doesn't necessarily need to depend on frameworks for Buck CLI.
+    # It's a hack to make generated project link to the frameworks.
+    + first_party_frameworks,
 )
 
 apple_binary(
@@ -271,9 +274,7 @@ apple_binary(
 
     deps = [
         "//Libraries/SwiftShared:SwiftShared",
-    ]
-    # See ":ExampleApp" for why this need to depend on frameworks.
-    + first_party_frameworks
+    ],
 )
 
 apple_bundle(
--- a/App/BUCK
+++ b/App/BUCK
@@ -1,5 +1,6 @@
 load("//Config:configs.bzl", "app_binary_configs", "library_configs", "watch_binary_configs", "message_binary_configs", "pretty", "info_plist_substitutions", "bundle_identifier", "DEVELOPMENT_LANGUAGE")
 load("//Config:buck_rule_macros.bzl", "apple_lib", "apple_test_lib", "apple_test_all")
+load("//Config:frameworks.bzl", "first_party_frameworks")
 
 apple_asset_catalog(
     name = "ExampleAppAssets",
@@ -31,6 +32,7 @@ first_party_library_dependencies = [
     "//Libraries/SwiftWithMLModel:SwiftWithMLModel",
     "//Libraries/SwiftWithPrecompiledDependency:SwiftWithPrecompiledDependency",
     "//Libraries/YetAnotherSwiftModule:YetAnotherSwiftModule",
+    "//Libraries/SwiftShared:SwiftShared",
 ]
 
 prebuilt_frameworks = [
@@ -73,6 +75,9 @@ apple_library(
     ]
     + first_party_library_dependencies
     + build_phase_scripts,
+    # The library doesn't necessarily need to depend on frameworks for Buck CLI.
+    # It's a hack to make generated project link to the frameworks.
+    + first_party_frameworks,
 )
 
 apple_binary(
@@ -116,7 +121,7 @@ xcode_postbuild_script(
 apple_test_all(
     name = "ExampleAppCITests",
     libraries = first_party_library_dependencies,
-    additional_tests = app_tests,
+    additional_tests = app_tests + first_party_frameworks,
     prebuilt_frameworks = prebuilt_frameworks,
 )
 
@@ -145,10 +150,10 @@ apple_bundle(
     product_name = "ExampleApp",
     info_plist = "Info.plist",
     info_plist_substitutions = info_plist_substitutions("ExampleApp"),
-    deps = [
+    deps = first_party_frameworks + [
         # For "#watch", https://buckbuild.com/rule/apple_bundle.html#deps
         ":ExampleWatchApp#watch",
-        ":ExampleMessageExtension"
+        ":ExampleMessageExtension",
     ]
     + prebuilt_frameworks,
 )
@@ -266,6 +271,12 @@ apple_binary(
         "-Xlinker",
         "@executable_path/../../Frameworks",
     ],
+
+    deps = [
+        "//Libraries/SwiftShared:SwiftShared",
+    ]
+    # See ":ExampleApp" for why this need to depend on frameworks.
+    + first_party_frameworks
 )
 
 apple_bundle(
--- a/Config/buck_rule_macros.bzl
+++ b/Config/buck_rule_macros.bzl
@@ -86,13 +86,14 @@ def apple_test_all(
         libraries = [],
         additional_tests = [],
         prebuilt_frameworks = [],
+        deps = [],
         **kwargs):
     ci_test_libraries = []
     for library in libraries:
         ci_test_libraries.append(ci_test_name(test_name(library)))
 
     apple_test_lib(
-        deps = ci_test_libraries + additional_tests + prebuilt_frameworks,
+        deps = ci_test_libraries + additional_tests + prebuilt_frameworks + deps,
         bundle_for_ci = False,
         **kwargs
     )
--- a/Config/buck_rule_macros.bzl
+++ b/Config/buck_rule_macros.bzl
@@ -1,4 +1,4 @@
-load("//Config:configs.bzl", "test_configs", "library_configs", "pod_library_configs", "DEVELOPMENT_LANGUAGE")
+load("//Config:configs.bzl", "test_configs", "library_configs", "framework_configs", "pod_library_configs", "DEVELOPMENT_LANGUAGE")
 
 # This is just a regular lib that was warnings not set to error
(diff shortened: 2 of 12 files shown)


Libraries/SwiftShared/BUCK
```diff
@@ -0,0 +1,3 @@
+load("//Config:buck_rule_macros.bzl", "first_party_framework")
+
+first_party_framework("SwiftShared")
```
bachand: Let's add a trailing newline for good measure.

Libraries/SwiftShared/Tests/SwiftSharedTests.swift
```diff
@@ -0,0 +1,7 @@
+import XCTest
+
+@testable import SwiftShared
+
+final class SwiftSharedTests: XCTestCase {
+
+}
```
bachand: Let's add a simple test here to prove that it can execute successfully in a test.

Config/configs.bzl
```diff
@@ -57,6 +57,21 @@ def binary_configs(name):
     }
     return configs
 
+def framework_configs(name):
+    framework_specific_config = {
+        "PRODUCT_BUNDLE_IDENTIFIER": bundle_identifier(name),
+        # Not sure why, but either adding this or removing -whole-module-optimization can make it compile
+        "SWIFT_COMPILATION_MODE": "wholemodule",
+        # Setting to NO allows us not to create a header file manually
```
bachand: I don't follow this comment. I am not sure what "manually" means.

Config/configs.bzl
```diff
@@ -57,6 +57,21 @@ def binary_configs(name):
     }
     return configs
 
+def framework_configs(name):
+    framework_specific_config = {
+        "PRODUCT_BUNDLE_IDENTIFIER": bundle_identifier(name),
+        # Not sure why, but either adding this or removing -whole-module-optimization can make it compile
+        "SWIFT_COMPILATION_MODE": "wholemodule",
```
bachand: I know we found this be the case elsewhere. Is it definitely the case here too?

Config/buck_rule_macros.bzl
```diff
@@ -201,6 +206,41 @@ def first_party_library(
         deps = [":" + name] + test_deps,
         **kwargs)
 
+# Use this macro to declare first-party frameworks which can be shared between bundles.
```
bachand: Let's use the word "dynamic" since a framework can be static or dynamic. I realize the "shared between bundles" should make that clear, but seems like it can't hurt to be more clear.

Config/buck_rule_macros.bzl
```diff
@@ -201,6 +206,41 @@ def first_party_library(
...
+# This macro is similar to first_party_library.
+# We should support more parameters in the future to accommodate different requirements.
+# - parameter name: The name of the apple_library created for the code in the Sources/ directory. The name will become the module name.
+def first_party_framework(name):
```
bachand: What do you think of calling this `first_party_dylib` since a framework can technically be static or dynamic?

App/BUCK
```diff
@@ -102,7 +103,8 @@ apple_bundle(
     deps = [
         # For "#watch", https://buckbuild.com/rule/apple_bundle.html#deps
         ":ExampleWatchApp#watch",
-        ":ExampleMessageExtension"
+        ":ExampleMessageExtension",
+        "//Libraries/SwiftShared:SwiftSharedFramework",
```
bachand: Can you help me understand why the `ExampleApp` library needs to depend on `//Libraries/SwiftShared:SwiftShared`, and the `ExampleAppBundle` needs to depend on `//Libraries/SwiftShared:SwiftSharedFramework`? Is this so that the framework gets embedded?

I'm confused why `ExampleMessageExtension` doesn't need to depend on `//Libraries/SwiftShared:SwiftSharedFramework`.

bachand (commented):
😮 

I had a few comments but overall this is looking great.

It seems like `first_party_framework` is much simpler than what you wrote out in the PR message. Is the PR message outdated?

Libraries/SwiftShared/Tests/SwiftSharedTests.swift
dfed: Might be good to have that test access a resource in the shared bundle. I can pair with you on this on Monday if you'd like 🙂

bachand: This is a great idea

qyang-nj:
I looked more closely to the output and found that the binary of generated project still links to a static library, although the framework is copied in the main app bundle.

I used `otool` to examine what dylibs are linked to the binary. The binary of Buck CLI build does link to `@rpath/SwiftShared.framework/SwiftShared`, while the binary of generated project does not.

image
(The screenshot of Buck CLI build. The binary of generated project doesn't have the highlighted line. )

Then I looked at the build phases. `SwiftShared` is built into both static library and framework. 😱
image

App/BUCK
qyang-nj: Answered in the PR description

Config/buck_rule_macros.bzl
qyang-nj: I think `framework` is more appropriate, since frameworks can contain libraries and resources. Also, I think we can just add another parameter to make this support static framework.  `binary = ":SwiftShared#static"`

Config/configs.bzl
qyang-nj: Nope. Removed.

Config/configs.bzl
qyang-nj: Never mind. This setting is not necessary.

Libraries/SwiftShared/Tests/SwiftSharedTests.swift
qyang-nj: Test case is added

Config/buck_rule_macros.bzl
bachand: Yeah, maybe we can just add a parameter to specify the link style.

App/BUCK
bachand: The new PR description is great. Thanks!

bachand (approved):

bachand:
Great work @qyang-nj and @dfed

Config/frameworks.bzl
```diff
@@ -0,0 +1,5 @@
+# This file contains all first party frameworks
+
+first_party_frameworks = [
+    "//Libraries/SwiftShared:SwiftSharedFramework",
+]
```
dfed: nit: newline at end of file

App/Tests/BUCK
```diff
@@ -1,13 +1,13 @@
...
+load("//Config:frameworks.bzl", "first_party_frameworks")
 
 apple_test_lib(
     name = "Tests",
     srcs = glob([
         "*.swift",
     ]),
     info_plist = "Info.plist",
-    deps = [
+    deps = first_party_frameworks + [
```
dfed: 🙏 thanks for cleaning this up!

App/BUCK
```diff
@@ -46,7 +48,10 @@ apple_library(
         "//Pods:CryptoSwift",
         ":ExampleAppResource",
         ":ExampleAppAssets",
-    ] + first_party_library_dependencies,
+    ] + first_party_library_dependencies
+    # The library doesn't necessarily need to depend on frameworks for Buck CLI.
+    # It's a hack to make generated project link to the frameworks.
```
dfed: Great comment 😄

Config/buck_rule_macros.bzl
```diff
@@ -15,6 +15,15 @@ def ci_test_name(name):
 
 DEFAULT_SWIFT_VERSION = "4.0"
 
+def shared_plist_info_substitutions(name):
+    substitutions = {
+        "CURRENT_PROJECT_VERSION": "1",
+        "DEVELOPMENT_LANGUAGE": "en-us",
```
dfed: at some point we should make `DEVELOPMENT_LANGUAGE` a constant. Another time though

Config/buck_rule_macros.bzl
```diff
@@ -201,6 +206,63 @@ def first_party_library(
...
+# This macro is similar to first_party_library, and we can support more parameters in the future.
+# - parameter name: The name of the apple_library created for the code in the Sources/ directory. The name will become the module name.
+# - parameter exported_headers: The public headers for the library
+def first_party_framework(
+        name,
+        exported_headers = [],
+        deps = [],
+        has_resource = False,
+        resource_files = None):
+    framework_name = "%sFramework" % name
```
dfed: Should we make a `framework_name` method like ewe do with `test_name`?

Config/buck_rule_macros.bzl
```diff
@@ -201,6 +206,63 @@ def first_party_library(
...
+        visibility = ["PUBLIC"],
+    )
+
+    apple_test_lib(
+        name = lib_test_name,
+        srcs = native.glob(["Tests/**/*.swift"]),
+        info_plist = "Tests/Info.plist",
+        deps = [
+            ":" + name,
+            ":" + name + "Framework"],
```
dfed: nit: `],` should be on its own line

Libraries/SwiftShared/BUCK
```diff
@@ -0,0 +1,8 @@
+load("//Config:buck_rule_macros.bzl", "first_party_framework")
+
+first_party_framework(
+    name = "SwiftShared",
+    exported_headers = glob(["Sources/**/*.h"]),
```
dfed: Do we need this line? Also, do we need a `SwiftShared.h`?

qyang-nj: We don't need that in this sample, but it mimics the situation in Airbnb code.

Config/buck_rule_macros.bzl
qyang-nj: I'd prefer to keep this way for now until we need to share this method.

bachand:
> We made the app binary (`//App:ExampleApp`) depends on both the `SwitfShared` and `SwfitSharedFramework`. As previously stated, depending on `SwiftSharedFramework` isn't necessary for Buck CLI, but it makes generated project links to the framework. Now, generated project links to both a static library (`libSwiftShared.a`) and a dynamic framework (`SwfitShared.framework`). Based on observation, the dynamic linking overrides the static linking, which makes generated project behaves the same way as Buck CLI build.

I am not sure that the overriding is happening properly. I actually think the code is being duplicated.

The only type in `SwiftShared` is `PublicSharedClass`. I wanted to verify that this type only appears in the `SwiftShared.framework` binary and not the main binary.

I first ran `nm` on the `SwiftShared` binary to see what the mangled names of these types are.

I ran `nm` on both the `ExampleApp` binary and the `SwiftShared` binary in the `SwiftShared.framework`.



They look like this in my build:

```
00000000000018a0 T _$S11SwiftShared06PublicB5ClassC16readFromResourceSSyF
00000000000017e0 T _$S11SwiftShared06PublicB5ClassCACycfC
0000000000001880 T _$S11SwiftShared06PublicB5ClassCACycfc
0000000000001f8c s _$S11SwiftShared06PublicB5ClassCMF
0000000000002210 b _$S11SwiftShared06PublicB5ClassCML
0000000000001830 T _$S11SwiftShared06PublicB5ClassCMa
00000000000021a0 d _$S11SwiftShared06PublicB5ClassCMf
0000000000002178 D _$S11SwiftShared06PublicB5ClassCMm
0000000000001f44 S _$S11SwiftShared06PublicB5ClassCMn
00000000000021b0 D _$S11SwiftShared06PublicB5ClassCN
0000000000001d60 T _$S11SwiftShared06PublicB5ClassCfD
0000000000001d40 T _$S11SwiftShared06PublicB5ClassCfd
0000000000001f1c s _$S11SwiftSharedMXM
```

I then `grep`'ed for those names in the `ExampleApp` binary, and they appear.
Screen Shot 2019-03-19 at 6 04 48 PM


This indicates that the code is being duplicated, which would make sense since we still see the main app linking to `libSwiftShared.a` in Xcode.

Screen Shot 2019-03-19 at 6 05 42 PM

bachand:
It seems like these symbols are being duplicated in a Buck CLI build as well.

I built the project with `make build`.

|  `ExampleApp` binary | `SwiftShared` binary |
| -------- | -------- |
| Screen Shot 2019-03-19 at 7 13 03 PM | Screen Shot 2019-03-19 at 7 11 35 PM |

qyang-nj:
Buck CLI build is expected. The `U` means undefined, which will be linked at runtime.

bachand:
@qyang-nj and I learned that https://github.com/airbnb/BuckSample/pull/53 caused the code duplication described in https://github.com/airbnb/BuckSample/pull/52#issuecomment-474643614

qyang-nj:
@bachand This PR should be working for both Buck CLI and generated Xcode project. I'd like you to take a closer look and validate it.

Config/buck_rule_macros.bzl
```diff
@@ -217,6 +219,71 @@ def first_party_library(
...
 
+# Use this macro to declare first-party frameworks (dylib and resource) which can be shared between bundles.
+# This macro is similar to first_party_library, and we can support more parameters in the future.
+# - parameter name: The name of the apple_library created for the code in the Sources/ directory. The name will become the module name.
+# - parameter exported_headers: The public headers for the library
+def first_party_framework(
+        name,
+        exported_headers = [],
+        deps = [],
+        has_resource = False,
```
jongsu253: Can we omit this line? 🤔

qyang-nj: Sure, if you don't have any resources.

jongsu253: I mean the code seems redundant, so, can we remove it? 😃

shepting (approved):

dfed:
Looking through open PRs where I'm tagged: should we close this?

PR #53, opened 2019-03-16
Use `-all_load` for binaries in Buck-generated project to work around Apple bug

bachand:
In https://github.com/airbnb/BuckSample/pull/38 I documented [Apple's bug](https://bugs.swift.org/browse/SR-6004) with Swift static libraries that can cause protocol conformances to be stripped under some circumstances.

This bug only manifests in builds produced by Buck-generated projects. It does not manifest in builds from the Buck CLI since those builds do not use Swift static libraries.

We have investigated how best to work around this bug in our Buck-generated projects.

We currently believe that adding `-all_load` to Apple binaries is the best approach.

`-force_load`ing specific static libraries can also work. However, the ergonomics of it aren't as good as you likely need to find and specify each library to `-force_load`. This can quickly become a game of whack-a-mole.

We are planning to ship to the App Store with Buck CLI builds so we won't encounter this issue in production. However, we want our developers to be able to use the Buck-generated projects with confidence and not worry if bugs they see in development are due to this Apple bug. Because we are only using Buck-generated project for development we also don't need to worry about bloating the binary size by adding `-all_load.` 

### Tags for SEO
- SR-6004

### DIFF
--- a/App/BUCK
+++ b/App/BUCK
@@ -1,4 +1,4 @@
-load("//Config:configs.bzl", "binary_configs", "library_configs", "watch_binary_configs", "message_binary_configs", "pretty", "info_plist_substitutions")
+load("//Config:configs.bzl", "app_binary_configs", "library_configs", "watch_binary_configs", "message_binary_configs", "pretty", "info_plist_substitutions")
 load("//Config:buck_rule_macros.bzl", "apple_lib", "apple_test_lib", "apple_test_all")
 
 apple_asset_catalog(
@@ -55,7 +55,7 @@ apple_binary(
         "//App:",
         "//App/...",
     ],
-    configs = binary_configs("ExampleApp"),
+    configs = app_binary_configs("ExampleApp"),
     swift_version = "4.0",
     srcs = [
         "BuckSupportFiles/Dummy.swift",
@@ -202,7 +202,7 @@ apple_binary(
         "MessageExtension/**/*.swift",
     ]),
     configs = message_binary_configs("ExampleApp.MessageExtension"),
-    
+
     # "-e _NSExtensionMain" tell linker this binary is app extension, so it won't fail due to missing _main
     # "-Xlinker -rpath -Xlinker @executable_path/../../Frameworks" tells the executable binary to
     # look for frameworks in ExampleApp.app/Frameworks instead of PlugIns, so that we don't need to have
--- a/App/ViewController.swift
+++ b/App/ViewController.swift
@@ -83,8 +83,9 @@ class ViewController: UIViewController {
     //
     // `-force_load`ing the module where the conformance exists also works.
     //
-    // This is tracked by https://bugs.swift.org/browse/SR-6004. This seems to indicate that
-    // `-all_load` will work too, but we haven't verified that.
+    // Finally, you can use `-all_load`. **That is the workaround that we are currently using.**
+    //
+    // This is tracked by https://bugs.swift.org/browse/SR-6004.
 //    let myObject = MyPublicClass()
     let myObject = MyFactory.myPublicObject()
     if (myObject as? MyPublicProtocol) == nil {
--- a/App/ViewController.swift
+++ b/App/ViewController.swift
@@ -67,26 +67,22 @@ class ViewController: UIViewController {
     // This line will crash if the assets from SwiftWithAssets haven't been bundled into the app
     _ = Catalog.buck.image
 
-    // Without an object explicitly typed as `MyPublicClass`, an instance of `MyPublicClass` won't
-    // exhibit conformance to `MyPublicProtocol`. `MyPublicClass` is defined in `Swift4`.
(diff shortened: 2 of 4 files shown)


Config/utils.bzl
```diff
@@ -0,0 +1,74 @@
+OTHER_LINKER_FLAGS_KEY = 'OTHER_LDFLAGS'
```
bachand: I've been meaning to bring these over for awhile. I brought over everything except `configs_without_key` which is not Skylark compliant.

`config_with_updated_linker_flags` was also not Skylark compliant, but I updated it since we need to use it.

In future PRs we could clean up some of our code by using these helpers.

App/BUCK
```diff
@@ -55,7 +55,7 @@ apple_binary(
         "//App:",
         "//App/...",
     ],
-    configs = binary_configs("ExampleApp"),
+    configs = app_binary_configs("ExampleApp"),
```
bachand: Given the additional `*_binary_configs` methods, it seemed worth clarifying this.

Config/configs.bzl
```diff
@@ -44,18 +50,15 @@ def library_configs():
...
 
-def binary_configs(name):
+def app_binary_configs(name):
     binary_specific_config = {
         "ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES": "YES",
         "DEVELOPMENT_LANGUAGE": "Swift",
         "PRODUCT_BUNDLE_IDENTIFIER": bundle_identifier(name),
     }
     binary_config = SHARED_CONFIGS + binary_specific_config
-    configs = {
```
bachand: Using `configs_with_config` results in a functionality change since the resulting `configs` dictionary will now include a `"Release"` key/value pair, but that seems like the right behavior anyway.

Config/utils.bzl
```diff
@@ -0,0 +1,74 @@
...
+                other_linker_flags)
+        return updated_configs
+
+# Either appends or assigns `other_linker_flags` to `config` under `config_key`.
+# Params:
+# - config: A dictionary of config names and their values
+# - additional_linker_flags: A string-representable value of additional linker flags
+# - config_key: The key to which to append or assign the additional linker flags
+def config_with_updated_linker_flags(config, other_linker_flags, config_key=OTHER_LINKER_FLAGS_KEY):
+    new_config = { }
```
bachand: This had to be rewritten from the following since dictionaries do not have a `copy()` method in Skylark.

```python
def config_with_updated_linker_flags(config, other_linker_flags, config_key=OTHER_LINKER_FLAGS_KEY):
    config_copy = config.copy()
    if config_key in config_copy:
        config_copy[config_key] += ' ' + other_linker_flags
    else:
        config_copy[config_key] = '$(inherited) ' + other_linker_flags
    return config_copy
```

App/ViewController.swift
```diff
@@ -83,8 +83,9 @@ class ViewController: UIViewController {
     //
     // `-force_load`ing the module where the conformance exists also works.
     //
-    // This is tracked by https://bugs.swift.org/browse/SR-6004. This seems to indicate that
-    // `-all_load` will work too, but we haven't verified that.
+    // Finally, you can use `-all_load`. **That is the workaround that we are currently using.**
+    //
+    // This is tracked by https://bugs.swift.org/browse/SR-6004.
 //    let myObject = MyPublicClass()
```
dfed: Should we remove this line and the part of the comment that refers to it above? Feels a little confusing to come into this project and see a "hey try this out to fix this issue" when the Xcode project will already be working as expected due to `-all_load`.

Config/configs.bzl
```diff
@@ -24,6 +26,10 @@ SHARED_CONFIGS = {
     "LD_RUNPATH_SEARCH_PATHS": "@executable_path/Frameworks", # To allow source files in binary
 }
 
+# Adding `-all_load` to our binaries works around https://bugs.swift.org/browse/SR-6004. See the
+# longer comment in `ViewController.swift` for more details.
+APPLE_BINARY_OTHER_LINKER_FLAGS = "-all_load"
```
dfed: Should we make this constant name specific to the use case? Or do we think we'll want to add more flags here?

If we think we'll want to add more flags, should we make this property an array instead of a string?

App/ViewController.swift
bachand: Good feedback: https://github.com/airbnb/BuckSample/pull/53/commits/90d5a53d118b2a63a27ea4cd6817f3267d402c6b

Config/configs.bzl
bachand: I think the simplest path forward is to un-generalize for now until we have a need to add more: https://github.com/airbnb/BuckSample/pull/53/commits/4fbace081ce80d6b490d65a233dfba38f3476a6f

bachand:
I regenerated and it's still working correctly. Will merge when CI checks complete.

Screen Shot 2019-03-18 at 12 43 33 PM

dfed:
Love the new commits!

PR #54, opened 2019-03-19
Add Build Phase Scripts

fdiaz:
This PR demonstrates how to add a Build Phase Script to our sample project both as pre-build and post-build.

This is how it looks in the generated project:
image

cc @shepting @dfed @bachand @qyang-nj

### DIFF
--- a/App/BUCK
+++ b/App/BUCK
@@ -29,6 +29,8 @@ first_party_library_dependencies = [
     "//Libraries/SwiftWithMLModel:SwiftWithMLModel",
 ]
 
+# Build Phase scripts need to be added as dependencies.
+# These only get executed when building with Xcode, not Buck.
 build_phase_scripts = [
     ":Hello_World",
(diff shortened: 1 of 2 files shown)


App/BUCK
```diff
@@ -46,7 +51,7 @@ apple_library(
         "//Pods:CryptoSwift",
         ":ExampleAppResource",
         ":ExampleAppAssets",
-    ] + first_party_library_dependencies,
+    ] + first_party_library_dependencies + build_phase_scripts,
```
dfed: Can we put `+ build_phase_scripts` on its own line? Or at least add a comment about what it is doing?

App/BUCK
```diff
@@ -65,6 +70,24 @@ apple_binary(
     ],
 )
 
+xcode_prebuild_script(
```
dfed: A comment re what this does would be lovely. Same for the script below.

PR #56, opened 2019-03-20
[WIP] Add `PromiseKit` 6.1.1. to example project [Option B]

bachand:
## Motivation

Right now the linker is complaining about duplicate symbols in `PromiseKit` when `-all_load` is enabled (see https://github.com/airbnb/BuckSample/pull/53).

## Possible approaches

1. Determine if there are changes we can make to the Buck configurations to resolve build errors.
2. Create a fork of `PromiseKit` 6.1.1 that resolves this issue.

## Current status

Right now the linker is complaining about duplicate symbols in `PromiseKit` when `-all_load` is enabled (see https://github.com/airbnb/BuckSample/pull/53).

## Reviewers

@shepting @qyang-nj @dfed @fdiaz @joeyg

------

**See Option A: https://github.com/airbnb/BuckSample/pull/55**

PR #55, opened 2019-03-20
[WIP] Add `PromiseKit` 6.1.1. to example project [Option A]

bachand:
## Motivation

Right now the linker is complaining about duplicate symbols in `PromiseKit` when `-all_load` is enabled (see https://github.com/airbnb/BuckSample/pull/53).

## Approach

This PR builds on top of https://github.com/airbnb/BuckSample/pull/52 to see if this errors are avoided when `PromiseKit` is a dynamic framework.

## Current status

- `make project` works.
- Builds errors in `PromiseKit`.
- Dependency on https://github.com/airbnb/BuckSample/pull/53

## Reviewers

@shepting @qyang-nj @dfed @fdiaz @joeyg

------

**See Option B: https://github.com/airbnb/BuckSample/pull/56**

PR #58, opened 2019-03-20
Add `PromiseKit` 6.1.1. to example project [Option E]

bachand:
The PR is the simplest solution for getting `PromiseKit` 6.1.1. integrated into `BuckSample`.

The issue we were having was the Buck-generated project wasn't building after we added `-all_load`. We were getting duplicate symbol errors.

It makes sense now why this was happening. Our Buck definition for `PromiseKit` was including too many files to compile.

The `PromiseKit` podspec doesn't list some files as sources.
Screen Shot 2019-03-20 at 1 25 32 PM

Here are the source being compiled by our manually managed project, which consumes `PromiseKit` via the podspec.
Screen Shot 2019-03-20 at 1 26 26 PM

Excluding two files gets the project to build.

Our manually managed project has 27 sources but this project has 26. The only file missing is `PromiseKit-dummy.m`, which I suspect is added by CocoaPods.

This is a replace for https://github.com/airbnb/BuckSample/pull/55 and https://github.com/airbnb/BuckSample/pull/56


Pods/BUCK
```diff
@@ -17,6 +17,26 @@ apple_third_party_lib(
...
 
+apple_third_party_lib(
+    name = "PromiseKit",
+    visibility = ["PUBLIC"],
+    exported_headers = glob([
+        "PromiseKit/**/*.h",
+    ], exclude = [
+        "PromiseKit/AnyPromise+Private.h",
+    ]),
+    headers = glob([
```
dfed: Since the `podspec` does not include any private headers, I don't think we need this line

Pods/BUCK
```diff
@@ -17,6 +17,26 @@ apple_third_party_lib(
...
+        "PromiseKit/**/*.h",
+    ]),
+    srcs = glob([
+        "PromiseKit/**/*.m",
+        "PromiseKit/**/*.swift",
+    ], exclude = [
```
dfed: 👏

Pods/BUCK
bachand: Good call. Removed.

Podfile
```diff
@@ -11,6 +11,8 @@ target 'BuckSample' do
   pod 'CryptoSwift'
   pod 'OHHTTPStubs'
   pod 'Bugsnag'
+  pod 'PromiseKit/CorePromise', '6.1.1'
```
bachand: There is nothing special about this version. It's just what we're using currently.

bachand:
I added code to use `PromiseKit` in `ViewController.swift` and verified that it works.

Screen Shot 2019-03-20 at 3 33 46 PM

Pods/BUCK
dfed: A comment on why we're excluding these things would be dope.

bachand: Great idea: https://github.com/airbnb/BuckSample/pull/58/commits/30bdd930fb85c282ed404027579be0959b2fce1d

PR #61, opened 2019-04-04
Update apple_bundle name

shepting:
This makes Buck commands a little more idiomatic. e.g. `buck build //App:ExampleApp`

 * Update the `apple_bundle` name to be `ExampleApp`
 * Update the `apple_library` to be named `ExampleAppLibrar`
 * Add Apple Watch, and other CI checks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 
-.PHONY : log install_buck build targets pods audit debug test xcode_tests clean project audit
+.PHONY : log install_buck build watch message targets pods audit debug test xcode_tests clean project audit
 
 # Use local version of Buck
 BUCK=tools/buck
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -16,15 +16,21 @@ update_cocoapods:
 	pod install
 
 build:
-	$(BUCK) build //App:ExampleAppBundle
+	$(BUCK) build //App:ExampleApp
+
+watch:
```
dfed: 👏

Makefile
```diff
@@ -1,5 +1,5 @@
 
-.PHONY : log install_buck build targets pods audit debug test xcode_tests clean project audit
+.PHONY : log install_buck build watch message targets pods audit debug test xcode_tests clean project audit
```
bachand: What does this line do? I’ve always wondered

Makefile
```diff
@@ -44,11 +50,13 @@ audit:
 	$(BUCK) audit rules Pods/BUCK > Config/Gen/Pods-BUCK.py
 
 clean:
```
bachand: Should we have `clean` depend on `killall`

Makefile
shepting: Make has trouble knowing when it should run a task again. Something about looking at file timestamps...

This seems to make it _always_ run that task.

Makefile
```diff
@@ -16,15 +16,21 @@ update_cocoapods:
...
 
 debug:
-	$(BUCK) install //App:ExampleAppBundle --run --simulator-name 'Phone: iPhone XS'
+	$(BUCK) install //App:ExampleApp --run --simulator-name 'Phone: iPhone XS'
 
 targets:
 	$(BUCK) targets //...
 
-ci: install_buck targets build test project xcode_tests
+ci: install_buck targets build test project xcode_tests watch message
```
shepting: This line runs extra builds on CI.

Makefile
shepting: I seem to remember being frustrated that some task killed my Xcode. I think that's why I made this change.

Makefile
bachand: 👍

PR #62, opened 2019-04-05
Remove .test extension on test_name

dfed:
We'd found that the `.test` extension can muck with Apple's bundle identifier expectations. @qyang-nj, do you remember what exactly the `.test` extension broke?


Config/buck_rule_macros.bzl
```diff
@@ -30,8 +30,6 @@ def apple_test_lib(
         frameworks = [],
         labels = [],
         **kwargs):
-    test_name = name + ".test"
```
dfed: As part of this PR, I'd love to include a test that'd break if we were to re-introduce this. @qyang-nj, any ideas?

bachand: Was it something about whether it treated it as a UI or unit test? Or maybe it was a macOS or iOS test? Something like that. I think Qing did a good job of explaining it in his internal PR IIRC

qyang-nj: It's about unit tests with a host application. In [Apple](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/testing_with_xcode/chapters/05-running_tests.html)'s word, it's called "app tests". 

> App tests run in the context of your app, allowing you to create tests which combine behavior that comes from the different classes, libraries/frameworks, and functional aspects of your app. Library tests exercise the classes and methods within a library or framework, independent of your app, to validate that they behave as the library’s specification requires.

dfed: Got it. So if we were to create a `HostedTest` target, before this PR it would have failed. Will add.

dfed: Actually, given the number of in-flight PRs, and our discussion in https://github.com/airbnb/BuckSample/pull/64#discussion_r272444327, I think we can hold off on adding that to a future PR.

Issue #18, opened 2019-01-18
ui_test are failing


gsabran:
`make ui_test` fails for me. Should they be part of the ci ?

```
make ui_test
/usr/local/Cellar/buck/2018.10.29.01/bin/buck test //App/UITests:UITests
FAILURE BuckSampleUITests -[BuckSampleUITests testExample]: <unknown>:0: failed: caught "NSInternalInconsistencyException", "No target application path specified via test configuration: <XCTestConfiguration: 0x600003368270>
```

dfed:
It [looks like] `xctool` isn't capable of running UI Tests. I'll take a task to investigate having `make ui_test` generate an Xcode project whose UI Test target is then run via `xcodebuild`

mrtom:
There's another (possibly related?) issue with the UI Tests. Even tho `test_host_app` seems to be set correctly in the BUCK file for UITests, if you view the UITests target in Xcode* you'll see the host app is set to none and can't be changed - when it should be set to `ExampleAppBundle`.

This indicates to me that Buck isn't creating a valid Xcode project file for this target for some reason, although I've failed to figure out how/why.

![screenshot 2019-03-07 at 16 54 47](https://user-images.githubusercontent.com/174462/53974557-b8539d80-40fa-11e9-9202-0058c54f0c48.png)

*you'll have to re-enable the UITests by adding them to the `app_tests` array in the main BUCK file to get them to show up back in Xcode.

dfed:
It kinda seems like neither `buck` nor `xctool` are set up to handle UI Tests 😬.

For what it's worth, internally we run screenshot tests using Buck's CLI using a host app, but we don't use Apple's UI testing framework to do so.

mrtom:
Thanks that's useful to know!

I wonder if @milend will get a ping if I mention him and comment on 'buck not set up to handle UI tests'?

qyang-nj:
The "Host Application is None" problem will be fixed by #62, it's because `.test` suffix breaks Buck hacks.

dfed:
@qyang-nj indeed! One problem down.

That said, UI tests (i.e. tests that run `XCUIApplication().launch()`) will still crash per my local testing. But we should be able to run tests in a host app, and grab screenshots in that host app without running `XCUIApplication().launch()`.

PR #64, opened 2019-04-05
Allow for running tests in a host app

dfed:
In order to run tests in a host app in Xcode, the unit test target and the host app must be part of the same `xcodeproj`, and therefore they must be defined in the same `BUCK` file.

While I was here, I also added the ability to specify a `destination_specifier` on a test library.

Note that this still does not get UI tests working. But this would allow for us to have a unit test that required a host app


App/BUCK
```diff
@@ -272,3 +270,39 @@ apple_resource(
...
+        ":ExampleAppLibrary",
+    ],
+)
+
+### Tests End ###
+
+### UI Tests Begin ###
+
+apple_test_lib(
+    name = "UITests",
```
qyang-nj: Unit tests with a host application is actually not UI tests. UI tests cannot access the code of the app, but unit tests can. I'm not sure if Buck supports UI tests.

dfed: I’m pretty sure Buck does not support UI tests per #18. But this was already here under the `App/UITests/` directory. We should probably remove this test (which is actually a UI test, and does fail), and instead replace it with something akin to our screenshot or accessibility tests.

thedavidharris: @dfed Buck at least recognizes the extension https://github.com/facebook/buck/blob/a6c7d7391a7a8e575c6372cac6690a240b5907ea/test/com/facebook/buck/apple/testdata/apple_test_xcuitest/BUCK.fixture#L64, `is_ui_test` isn't documented though. Haven't tried it.

dfed: TIL! Thanks for surfacing. We'll give it a shot.

fdiaz: I did try using `is_ui_test` without success in our app, but hey, we could try again here?

dfed: Shockingly, that worked: https://github.com/airbnb/BuckSample/pull/70. Thanks!!

PR #66, opened 2019-04-05
Add prebuilt framework AFNetworking

dfed:
Resolves https://github.com/airbnb/BuckSample/issues/43

cc @thedavidharris


App/BUCK
```diff
@@ -27,6 +27,14 @@ first_party_library_dependencies = [
     "//Libraries/SwiftReliesOnCXX:SwiftReliesOnCXX",
     "//Libraries/SwiftWithAssets:SwiftWithAssets",
     "//Libraries/SwiftWithMLModel:SwiftWithMLModel",
+    "//Libraries/SwiftWithPrecompiledDependency:SwiftWithPrecompiledDependency",
+]
+
+prebuilt_frameworks = [
+    "//Carthage:AFNetworking",
+
+    # Hack to enable project generation to work for the Carthage BUCK file which otherwise only has `prebuilt_apple_framework`s.
```
thedavidharris: Noticed this as well, was pretty weird

App/BUCK
```diff
@@ -272,3 +279,40 @@ apple_resource(
...
+apple_test_lib(
+    name = "Tests",
+    srcs = glob([
+        "Tests/*.swift",
+    ]),
+    info_plist = "Tests/Info.plist",
+    deps = [
+        ":ExampleAppLibrary",
+    ]
+    + prebuilt_frameworks,
```
thedavidharris: Yeah this was kind of where we've landed as well. It's pretty unsatisfying, since Buck can pick up the transitive dependencies for built targets but it won't for prebuilt targets. In some cases we've got a prebuilt framework wrapped by a framework we have to extend that, and the everything up the dependency chain needs to explicitly include that prebuilt target all the way up the tree.

dfed: Yup. This is clearly not ideal. I'd love for this to be fixed in Buck, but for now we're working around it.

dfed:
@thedavidharris good call re trying `tools/buck test //Libraries/SwiftWithPrecompiledDependency:SwiftWithPrecompiledDependency`. As is, that command crashes because `AFNetworking` isn't found 😫. I'll fix.

Libraries/SwiftWithPrecompiledDependency/BUCK
```diff
@@ -0,0 +1,12 @@
+load("//Config:buck_rule_macros.bzl", "first_party_library")
+
+first_party_library(
+    name = "SwiftWithPrecompiledDependency",
+    deps = [
+        "//Carthage:AFNetworking",
+    ],
+    test_deps = [
+        # Explicitly add the precompiled library to the test's dependencies to prevent a crash when running this test target in CLI.
+        "//Carthage:AFNetworking",
```
dfed: @thedavidharris here was the fix for `tools/buck test //Libraries/SwiftWithPrecompiledDependency:SwiftWithPrecompiledDependency`. I don't love this 😬.

thedavidharris: Yep, that was pretty much the whole issue. Super annoying to have to define this all the way throughout the tree. It only happens with dynamic frameworks at least, but nobody is really shipping precompiled static frameworks. Unfortunately most of our projects rely on a few closed-source dynamic libraries from vendors, so it ends up being a bit pervasive.

In a Carthage project we've made this a little easier on test targets by just having the Runpath Search Paths point to the Carthage/Build/iOS directory, which is hacky but also reduces the size of the xctest bundles a lot. The top level .app targets still need the `carthage-copy` script as that directory won't exist as a runpath outside the simulators.

Might be easier to find something from CocoaPods too that ships a precompiled dynamic framework too if you want to keep this to one dependency manager.

dfed: The other option is to add `prebuilt_frameworks` to `test_deps` down in `first_party_library`. This isn't great, since it adds dependencies that aren't real. But... it would prevent needing to set it everywhere. Thoughts?

thedavidharris:
Interestingly enough it looks like this did work as expected at one point: https://github.com/facebook/buck/pull/1971 

But it was reverted, didn't really say why: https://github.com/facebook/buck/commit/3f3e7d586ef8cbf96fc34842d3963042ce587ca6

Libraries/SwiftWithPrecompiledDependency/BUCK
thedavidharris: It would save you a line I guess to help you define it in the `test_deps` automatically, but then anything that has an `import SwiftWithPrecompiledDependency` is still going to have to add `Carthage:AFNetworking` as a `test_dep`, but anything without that import you probably don't want it to be linked.

dfed: Good point. I guess what I’d recommend then is:

1. Create a protocol abstraction for your prebuilt library
2. Inject an instance that conforms to the protocol into your modules
3. Only link the prebuilt library from the main app target

It’s a bit of a pain to get there (likely an understatement), but the benefits are a simpler Buck structure, and a more testable codebase.

dfed:
> Were the changes to Pods/PromiseKit/README.md intentional?

Absolutely not 🤦‍♂️. Rogue find/replace. Will fix!

bachand:
I can’t say I fully followed this by just reading the PR and issue fairly quickly (I think I’d have to play around with it and I’m AFK) but it generally looks reasonable and I love the documentation in Github and in the code 👍

@thedavidharris could you approve this when you think it’s ready to merge? Thanks for helping us make `BuckSample` better 😊

Libraries/SwiftWithPrecompiledDependency/BUCK
thedavidharris: Yeah, we actually have to do that a lot on our core sets of apps. Different map providers, different sets of core frameworks as "plugins," it's doable but not always worth it.

Wondering if looking into the rpaths like https://github.com/airbnb/BuckSample/pull/52/files#diff-85ae0899ab92ae7735513ea90991ee32R231 does can help look this up. I know for a fact that `tools/buck test //Libraries/SwiftWithPrecompiledDependency:SwiftWithPrecompiledDependency` if you set the runpath search paths to include `Carthage/Build/iOS`, I'm just not sure what the implications of that are.

dfed: Oh fascinating! I wonder if we can use `exported_linker_flags` here to propagate the necessary linker flags up the dependency chain 🤔

dfed: A quick attempt didn't seem to have the desired affect – looks like the prebuilt framework isn't copied into a consuming executable unless it is an explicit dependency 😕. Let's investigate this on a later pass – I'd like to land this as-is, since it represents our current understanding. I'm happy to file an issue to improve upon this setup for prebuilt frameworks, because I absolutely agree the solution we're providing is less-than-ideal.

cc @qyang-nj who understands this stuff _way_ better than I do.

qyang-nj: I'm now sure if I'm following the problem, but `apple_bundle` needs to depend on those frameworks. That's where the frameworks are copied into the bundle.

dfed:
@thedavidharris would love your blessing (or improvement suggestions!) before this gets merged. This PR documents not the world we want, but it does document the world we have. Happy to file a new issue stating that we want to do this better, but I think having this as an example of our current (less-than-great) workaround is valuable.

thedavidharris:
Yeah it's good to me, definitely documents the current scenario well. Thanks for the undertaking!

Libraries/SwiftWithPrecompiledDependency/BUCK
thedavidharris: You're right @qyang-nj, the tricky part is that if a subdependency has a dependency on a prebuilt framework, Buck doesn't pick that up transitively. The quickest way to see this is for test targets: SwiftWithPrecompiledDependency has to define AFNetworking as a dependency for both deps and test_deps. Non-prebuilt frameworks don't work this way.

Xcode works similarly in that you need either a `carthage-copy` or to follow https://github.com/Carthage/Carthage#adding-frameworks-to-unit-tests-or-a-framework to get them into the right runpath location; or, what we've done on our projects is just point to the prebuilt framework directory in Runpath Search Paths. A lot of our project runs out of Carthage, so this reduces the XCTest bundle size IMMENSELY. We actually kept running out of space on our CI machines before we did this.

This doesn't play nice for modular structure though, as every target that has a prebuilt framework will need to set that rpath, and then it won't work when you build to a device because that runpath isn't that, from what I gather.

dfed: > A lot of our project runs out of Carthage, so this reduces the XCTest bundle size IMMENSELY

Fascinating! I understand why frameworks would have lowered the size of your `xctest` bundle, but wouldn't the size from the frameworks have moved 1:1 from the `xctest` bundle to the prebuilt framework? Or is the issue that you had multiple `xctest` bundles that relied on the same frameworks?

I'm curious if y'all see a similar size savings with [apple_test_all](https://github.com/airbnb/BuckSample/blob/30442b4e9e2849a2caa618eeb02d3c321bdae1a7/Config/buck_rule_macros.bzl#L83), since `apple_test_all` creates a single target with all of your unit test code, rather than creating a `xctest` bundle per target.

thedavidharris: Multiple XCTest bundles that all relied on the same frameworks. Each framework would then embed dependency A.framework in the `@rpath/Frameworks`, which would be a runtime location different for each XCTest bundle. But if you point the runpaths to the folder where the framework already is on the filesystem, then you never have to copy it into the XCTest bundle.

`apple_test_all` is honestly the biggest why I want to move these projects to Buck. Getting everything into one XCTest bundle would be hugely time and space efficient instead of relying on one bundle + test host (where needed) per framework target.

PR #70, opened 2019-04-05
Get UITests working with Buck

### LINKED ISSUE #18: ui_test are failing

dfed:
That was surprisingly easy. Thank you @thedavidharris for the idea!

This PR resolves #18.

### DIFF
--- a/App/BUCK
+++ b/App/BUCK
@@ -16,6 +16,7 @@ apple_resource(
 
 app_tests = [
     ":Tests",
+    ":UITests",
 ]
 
 first_party_library_dependencies = [
@@ -299,6 +300,7 @@ apple_test_lib(
         "UITests/*.swift",
     ]),
     info_plist = "UITests/Info.plist",
+    is_ui_test = True,
     labels = ['ui'],
     deps = [
         "//Libraries/ASwiftModule:ASwiftModule",
--- a/Config/buck_rule_macros.bzl
+++ b/Config/buck_rule_macros.bzl
@@ -27,6 +27,7 @@ def apple_test_lib(
         info_plist_substitutions = {},
         test_host_app = None,
         run_test_separately = False,
+        is_ui_test = False,
         frameworks = [],
         labels = [],
         destination_specifier = None,
@@ -61,6 +62,7 @@ def apple_test_lib(
         info_plist_substitutions = substitutions,
         destination_specifier = destination_specifier,
         test_host_app = test_host_app,
+        is_ui_test = is_ui_test,
         run_test_separately = run_test_separately,
         configs = test_configs(name),
         frameworks = [
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ debug:
 targets:
 	$(BUCK) targets //...
 
-ci: install_buck targets build test project xcode_tests watch message
+ci: install_buck targets build test ui_test project xcode_tests watch message
 	echo "Done"
 
 test:


gsabran:
@dfed could you share more on why they were removed (https://github.com/airbnb/BuckSample/commit/4c2d68d66d27d63450b6f90e0fb165b0d16114b4)

I'm trying to figure out why `tools/buck test //App:XCUITests` succeeds but runs no test.

dfed:
I had assumed it was because we never wrote any UI tests. The UI test file is just the Xcode template. If you write a UI test, does the test get found?

(Note: this is my first time dealing with UI tests, so it is quite possible I’m making an incorrect assumption somewhere along the way)

gsabran:
Should not https://github.com/airbnb/BuckSample/blob/master/App/XCUITests/BuckSampleUITests.swift#L28 count as one test? It shows up in Xcode and you can run it (it just installs and launches the app). I'm also very confused about UI tests and what Buck can or cannot handle.

Screen Shot 2019-04-08 at 2 49 53 PM

bogo:
That's a great start – at least now we can run the XCUITests directly from Xcode!

Looking at Buck's [test fixuture](https://github.com/facebook/buck/blob/a6c7d7391a7a8e575c6372cac6690a240b5907ea/test/com/facebook/buck/apple/testdata/apple_test_xcuitest/BUCK.fixture) I believe that `ui_test_target_app` is the missing parameter to get further on this – unfortunately, adding it causes an abort in the test phase with `xctool failed with exit code 4: ERROR: Unexpected action: -uiTest`.

Per [`RunTestActions.m`](https://github.com/facebook/xctool/blob/fe6fb5dfab112cfbb0a2fbda1f6a528abbe0d272/xctool/xctool/RunTestsAction.m#L199) `xctool` does not support `-uiTest` parameter, but Buck still [adds it](https://github.com/facebook/buck/blob/0cf9e77a4ebbadb5b17fcedfa36cfb8b2d6bcf85/src/com/facebook/buck/apple/XctoolRunTestsStep.java#L511). @ExtremeMan's experimental branch of `xctool`, as mentioned in xctool#534, does support `-uiTest` action, but I did not experiment with it.

ExtremeMan:
Unfortunately xctool doesn't currently support running UI tests. I will be happy to review PR for that but all my attempts to add such a support failed. It appears to be much more complex than other use cases we support.

bogo:
Thanks for the input @ExtremeMan!

It seems that there is something akin to `-uiTest` support in [facebook/FBSimulatorControl/fbxctest](https://github.com/facebook/FBSimulatorControl/blob/ff8691e4d4235a98f4a24f18074783a42943b710/fbxctest/FBXCTestKit/Configuration/FBXCTestCommandLine.m#L107). Perhaps that's what the test fixture is using in lieu of `xctool` for the UI tests alone? Will experiment and report back!

shepting:
@bogo We use `EarlGrey` for our UI Tests. That might be a handy thing to get set up in this sample app for others who use the same framework.

bogo:
Looked into this further - turns out `fbxctest` indeed is used instead `xctool` for those specific test fixtures in Buck. By both running the `XCUITests.xctest` and trying to run it via `fbxctest`, I've been making some progress – for now, I'm somewhat stuck on the fact that both approaches quit on me with:

```Error Domain=com.facebook.FBControlCore Code=0 "Timed out after 20.000000 seconds waiting for Connection to happen, _XCT_testBundleReadyWithProtocolVersion:minimumVersion: has not been called yet" UserInfo={NSLocalizedDescription=Timed out after 20.000000 seconds waiting for Connection to happen, _XCT_testBundleReadyWithProtocolVersion:minimumVersion: has not been called yet}```

...which I assume means that there is some missing XCUITest behavior on the app level, will investigate further.

Issue #69, opened 2019-04-05
Add `UnitTestsWithHostApp` target so as to show examples of all test types


bachand:
As far as we know, Buck does not support UI tests cc: @dfed @qyang-nj .

Right now we have a target named `UITests`. In reality it sounds like this may be `UnitTestsWithHostedApp`.

I wonder if we should rename our test targets to be more clear. As a strawman, what if we did the following rename?

`:Tests` => `:UnitTests`
`:UITests` => `:UnitTestsWithHostedApp`

dfed:
If we rename `:UITests` => `:UnitTestsWithHostedApp`, we should remove the `XCUIApplication().launch()` line from our `UITests` target. Since that is the line that actually makes our `UITests` target a `XCUITest` (as far as I can tell).

Fwiw, I do think we should do both of these things per https://github.com/airbnb/BuckSample/pull/64#discussion_r272444327

qyang-nj:
I like this naming a lot. Just one thing, it should be `WithHostApp`, not `Hosted`. The app hosts the tests, not the other way around.

dfed:
Per https://github.com/airbnb/BuckSample/pull/70, I no longer think we need the naming changes.

bachand:
Wow nice!

I find the distinctions between these many types of tests very confusing.

I wonder if we should consider adding a third test target now: `UnitTestsWithHostApp`?

Then `BuckSample` could demo all of the different types of tests that you can run.

Even better we could add a comment above each test explaining why you would want to use one vs the other.

If you like that idea @dfed and @qyang-nj I can close this issue and open another that is more descriptive based on this new information.

dfed:
> I wonder if we should consider adding a third test target now: UnitTestsWithHostApp

Absolutely. I like this idea!

> If you like that idea @dfed and @qyang-nj I can close this issue and open another that is more descriptive based on this new information.

I’m happy to repurpose this issue rather than creating a new one

bachand:
Cool I updated the title.

If we do this we should probably rename the existing tests target to be `UnitTests`

dfed:
@bachand see #71 😉

bachand:
Hurray!

PR #71, opened 2019-04-05
Improve test structure

dfed:
This PR resolves https://github.com/airbnb/BuckSample/issues/69, and also generally improves how we declare tests.


bachand:
😍 this is great! Small nit is that it could be cool to add rules for each of them to our `Makefile`. AFAICT it seems like these tests are run in CI though so that’s great.

Thanks @dfed 🙏

dfed:
I was tempted to _remove_ `ui_test` from our Makefile, since all the tests will be run with `make test` at this point. I think these were separated out before because `ui_test` wasn't working, but I'm not sure.

bachand:
> I was tempted to _remove_ `ui_test` from our Makefile

I support this. What I mostly find confusing is that we have a rule for one type of test and not the others.

dfed:
And here's the PR: https://github.com/airbnb/BuckSample/pull/72

PR #73, opened 2019-04-06
add c++ library

gsabran:
Add a library that runs some c++ code, wrapped in ObjC.


Config/buck_rule_macros.bzl
```diff
@@ -136,6 +136,7 @@ def apple_lib(
 def first_party_library(
         name,
         has_objective_c = False,
+        has_cpp = False,
```
qyang-nj: From the parameters, it seems that the library can have any combination of swift, objc and cpp source code. Does this work for a case that contains both swift and cpp? If so, can you add a test case?

Libraries/Cpp1/Cpp1.modulemap
```diff
@@ -0,0 +1,6 @@
+framework module Objc1 {
```
dfed: I'm surprised you need to manually create a module map and an umbrella header. Does Buck not generate these for you due to the module having C++ in it?

Libraries/Cpp1/Sources/CppClass1.h
```diff
@@ -0,0 +1,5 @@
+#import <Foundation/Foundation.h>
+
+@interface Cpp1 : NSObject
```
dfed: Can we add a comment to this file saying that it is a header file for an Objective-C++ class?

Libraries/Cpp1/Sources/util.cpp
```diff
@@ -0,0 +1,7 @@
+#include "util.hpp"
+//Num::Num() : num(0) { }
```
dfed: let's remove this commentated out line if it isn't necessary

Libraries/Cpp1/Tests/Info.plist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dfed: If you rebase `master` I don't think you'll need this file. I obviated the need for it [yesterday](https://github.com/airbnb/BuckSample/pull/71/files#diff-85ae0899ab92ae7735513ea90991ee32R50).

Config/buck_rule_macros.bzl
dfed: By test case, we mean adding a `swift` file to this new module? Or something else?

Libraries/Cpp1/Sources/CppClass1.mm
```diff
@@ -8,4 +8,8 @@ - (NSInteger)add:(NSInteger)a with:(NSInteger)b {
     
     return a_cpp.getNum() + b_cpp.getNum();
 }
+
++ (NSInteger)add:(NSInteger)a with:(NSInteger)b {
+    return 0;//addition(a, b);
```
dfed: I'm guessing this change wasn't supposed to be committed?

Libraries/Cpp1/Sources/MyBridgedSwiftClassInMixedModule.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+public class MyBridgedSwiftClassInCxxModule: NSObject {
+
+  public func findObjectiveCInstance() -> Any {
+    // Prove that Swift can find Objective C code in this module without a bridging header.
```
dfed: Nice!!

dfed: Though, quick nit: I think there is a bridging header, but it's generated by BUCK. It's because we have gathered the `.h` files into the `exported_headers`. This happens under the hood in `first_party_library`

Config/buck_rule_macros.bzl
gsabran: I added a swift file to this module

Libraries/Cpp1/Sources/MyBridgedSwiftClassInMixedModule.swift
gsabran: That was mainly a copy paste from https://github.com/airbnb/BuckSample/blob/master/Libraries/ObjcAndSwift/Sources/MyBridgedSwiftClassInMixedModule.swift#L6 I'll remove the comment.

Libraries/Cpp1/Cpp1.modulemap
gsabran: You're right. I was learning by imitation. It seems that this one can be removed as well: https://github.com/airbnb/BuckSample/tree/master/Libraries/Objc1

gsabran:
I'm not sure what to think of [cxx_library](https://buckbuild.com/rule/cxx_library.html) vs [apple_library](https://buckbuild.com/rule/apple_library.html). This PR is using the later at the moment through the `apple_cxx_lib` macro.

Libraries/Cpp1/Tests/Info.plist
gsabran: nice!

Libraries/Cpp1/Sources/CppClass1.mm
gsabran: :)

Libraries/Cpp1/Cpp1.modulemap
dfed: Good find! I'll put up a PR to clean up Objc1.

Libraries/Cpp1/Tests/Info.plist
dfed: To give credit where credit is due, I was just pushing upstream what @qyang-nj and @fdiaz had done internally 🙏

Libraries/Cpp1/Sources/MyBridgedSwiftClassInMixedModule.swift
dfed: Darnit. That's my comment too! I'll clean that one up.

dfed:
I think `apple_library` is the best bet – I'd guess that `cxx_library` wouldn't handle things like bridging headers for us.

Libraries/Cpp1/Sources/MyBridgedSwiftClassInMixedModule.swift
dfed: Actually, now I'm remembering how this works: bridging headers aren't needed in libraries (since the umbrella header acts as a bridging header), but they _are_ needed in executable targets (which do not have umbrella headers). I'll make this comment more explicit.

gsabran:
I don't think I've access to merge. Feel free to go ahead. Thanks for the reviews!

dfed:
Welp, forgot we had that restriction. Merging!

PR #79, opened 2019-04-10
Make sure all dependencies are explicit, and used. Alphabetize.

dfed:
`YetAnotherSwiftModule` was transitively included from `SecondSwiftModule`, so the fact that we were importing it didn't cause a build error. However, for `YetAnotherSwiftModule`'s tests to be executed, they need to be in `first_party_library_dependencies`.

While I was here, I also alphabetized a few things. And updated our `From CryptoSwift` code to actually use `CryptoSwift`.


bachand:
Nice!

PR #80, opened 2019-04-10
Enable code coverage for Buck build

qyang-nj:
This PR demonstrates code coverage with Buck CLI build.

[This article](http://clang.llvm.org/docs/SourceBasedCodeCoverage.html) explained how  to enable code coverage in detail.  Besides passing `-fprofile-instr-generate -fcoverage-mapping` to the obj-c compiler, we also need to pass `-profile-generate -profile-coverage-mapping` to the swift compiler and pass `-fprofile-instr-generate` to the linker.

The difficult part is setting environment variable LLVM_PROFILE_FILE to the simulator, where the tests are actually running. We need to do this through `buck` to `xctool`. Although `buck test` has an option `—test-runner-env`, but it doesn’t seem to be working. It turns out that `xctool` only accepts environment variable with `XCTOOL_TEST_ENV_` prefix, so it should be `—test-runner-env XCTOOL_TEST_ENV_LLVM_PROFILE_FILE=path/to/profraw`. This isn’t documented anywhere, besides [a comment](https://github.com/facebook/buck/blob/0cf9e77a4ebbadb5b17fcedfa36cfb8b2d6bcf85/src/com/facebook/buck/apple/XctoolRunTestsStep.java#L218) in buck source code.

After testing, we merge  `profraw`  files into  `Coverage.profdata` by `llvm-profdata`. Once we have `Coverage.profdata` and the app binary, we can use[`llvm-cov`](https://llvm.org/docs/CommandGuide/llvm-cov.html#llvm-cov-report) to generate the report, show coverage in the code, etc.

_showing the report_
Screen Shot 2019-04-09 at 10 12 30 PM

_showing uncovered area_
Screen Shot 2019-04-09 at 10 16 23 PM

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -41,3 +41,4 @@
   code_coverage_cxxflags = -fprofile-instr-generate -fcoverage-mapping
   code_coverage_ldflags = -fprofile-instr-generate
   code_coverage_swift_compiler_flags = -profile-generate -profile-coverage-mapping
+  
--- a/.buckconfig
+++ b/.buckconfig
@@ -1,14 +1,14 @@
 [cxx]
   default_platform = iphonesimulator-x86_64
-  cflags = -g -fmodules -fobjc-arc -D BUCK -w
-  cxxflags = -fobjc-arc -std=c++14 -D DEBUG -g
+  cflags = -g -fmodules -fobjc-arc -D BUCK -w $(config custom.other_cflags)
+  cxxflags = -fobjc-arc -std=c++14 -D DEBUG -g $(config custom.other_cxxflags)
   combined_preprocess_and_compile = true
   pch_enabled = false
-  ldflags = -Xlinker -objc_abi_version -Xlinker 2 -fobjc-arc -fobjc-link-runtime
+  ldflags = -Xlinker -objc_abi_version -Xlinker 2 -fobjc-arc -fobjc-link-runtime $(config custom.other_cxxflags)
 
 [swift]
   version = 4.0
-  compiler_flags = -DBUCK -enable-testing -g -Onone -whole-module-optimization
+  compiler_flags = -DBUCK -enable-testing -g -Onone -whole-module-optimization $(config custom.other_swift_compiler_flags)
   use_filelist = true
 
 [apple]
@@ -33,7 +33,11 @@
   ignore = tools, \
            .git, \
 
-
 [build]
   threads = 4
 
+[custom]
+  code_coverage_cflags = -fprofile-instr-generate -fcoverage-mapping
+  code_coverage_cxxflags = -fprofile-instr-generate -fcoverage-mapping
+  code_coverage_ldflags = -fprofile-instr-generate
+  code_coverage_swift_compiler_flags = -profile-generate -profile-coverage-mapping
--- a/Makefile
+++ b/Makefile
@@ -33,8 +33,18 @@ targets:
 ci: install_buck targets build test project xcode_tests watch message
 	echo "Done"
 
-test:
-	$(BUCK) test //App:ExampleAppCITests --test-runner-env FOO=BAR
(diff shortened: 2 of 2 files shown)


.buckconfig
```diff
@@ -33,7 +33,11 @@
...
 
-
 [build]
   threads = 4
 
+[custom]
+  code_coverage_cflags = -fprofile-instr-generate -fcoverage-mapping
+  code_coverage_cxxflags = -fprofile-instr-generate -fcoverage-mapping
+  code_coverage_ldflags = -fprofile-instr-generate
+  code_coverage_swift_compiler_flags = -profile-generate -profile-coverage-mapping
```
dfed: newline at end of file?

.buckconfig
```diff
@@ -1,14 +1,14 @@
 [cxx]
   default_platform = iphonesimulator-x86_64
-  cflags = -g -fmodules -fobjc-arc -D BUCK -w
-  cxxflags = -fobjc-arc -std=c++14 -D DEBUG -g
+  cflags = -g -fmodules -fobjc-arc -D BUCK -w $(config custom.other_cflags)
+  cxxflags = -fobjc-arc -std=c++14 -D DEBUG -g $(config custom.other_cxxflags)
```
dfed: do we need these flags here? Or do we only want these flags on our test targets/libraries? Is there a downside to always using these flags?

qyang-nj: Those flags will make compiler insert instrument code into the binary. When the binary is executing, those code will record/output if they are reached. This is generally how code coverage works. Since instrument code is added, we only want those when we run tests.

dfed: That makes sense, but, ugh. I'm wondering if this will affect local build times at all in Xcode, since these flags will be added to our Xcode project even though we only need these flags for CLI.

qyang-nj: Oh, no. It's only added in `make test` by `--config custom.other_cflags="$(config custom.code_coverage_cflags)"`. That being said, if you don't run `make test`, those will be empty.

dfed: I misunderstood! In that case: awesome 😄

mgrebenets:
Looks awesome, can't wait to try it out in our setup 🎉

I've got a question about the `ldflags` part:

> -Xlinker -objc_abi_version -Xlinker 2 -fobjc-arc -fobjc-link-runtime

are the first 2 flags new to Swift 5 and ABBI?
the `.buckconfig` I got ages ago never had `-fobjc-arc` and `-fobjc-link-runtime`.
The last one is about linking Swift stdlib, is it?


Also interested in use of `-Onone -whole-module-optimization` together.
Is it on purpose?
How does sample project manage Debug vs Release configurations?

qyang-nj:
1. I'm not certain why we have `-objc_abi_version 2`, but they are definitely not new to Swift 5.
2. `-fobjc-arc` just turns on ARC.
3. My understanding is that `-whole-module-optimization` affects how the code are built (primary file vs whole module), and big O optimizations (-O, -Osize, -Onone) affects the output binary, so they are not actually conflict.
4. I don't think BuckSample has Release build yet. Internally, we manage it by passing different flags through `--config` to buck.

mgrebenets:
Yep, makes sense. 
I assumed the Xcode's `SWIFT_COMPILATION_MODE` default `singlefile` maps to no flags, while `wholemodule` maps to `-whole-module-optimization`. Putting all files into one whole module first allows some extra optimizations if I understand it correctly.

Apple says this in their doco (https://help.apple.com/xcode/mac/10.2/#/itcaec37c2a6)

image

So I was thinking using `-whole-module-optimization` and `-Onone` together, would make the compiler compile sources in whole module mode, but apply no optimizations. 
Not sure if whole module vs single file has compile speed impacts when it comes to buck.

PR #82, opened 2019-04-14
Remove the unnecessary, extraneous xctool

bogo:
There are two copies of `xctool` in BuckSample. The one located in `vendor` is the one actually used in `.buckconfig`. This PR removes the unnecessary copy of `xctool`. It might be useful to move the path for the local installation of Buck as well.


bogo:
I don't have a write access to the repo, @qyang-nj. Would you mind landing this for me? Thanks!

qyang-nj:
I just checked the commit history. `tools/xctool` was updated by @shepting six months ago, and `vender/xctool` was updated by @dfed a moth ago. Honestly I don't know the difference, but I think we should keep the most up-to-date one.

bogo:
Yep, I initially removed the `tools/xctool` one, but then CI was unhappy due to `tools` being an empty folder. (so `make install_buck` was unable to succeed`) The second commit in the chain moved `vendor/xctool` to `tools/xctool`. I'm pretty indifferent – just gotta settle on `vendor` vs `tools` for all third party binaries. ;-)

qyang-nj:
Perfect, thanks!

bachand:
Seems like long-term we should use `vendor` for any third-party binaries.

bachand:
(But `tools` seems fine for now as we evolve things 👍)

PR #83, opened 2019-04-14
Add support for running UI tests from CLI

bogo:
This PR builds on #70 and adds support for running XCUITests directly from CLI. 

The actual reason for not being able to run UI tests from CLI was two-fold:
 - the original PR is missing a `ui_test_target_app` parameter in the `XCUITests` test library definition,
 - Buck's default iOS test-runner, `xctool`, does not recognize the `-uiTest` parameter that Buck passes to it.

By reading the sources of Buck's tests, I found out that the test cases for `XCUITest` support actually subsitute `apple.xctool_path` for a path to a different Facebook tool, [`fbxctest`](https://github.com/facebook/FBSimulatorControl/tree/master/fbxctest). By adding the `fbxctest` build from Buck's test suite and adding the missing `libShimulator.dylib` and `libMaculator.dylib` (also from [facebook/FBSimulatorControl](https://github.com/facebook/FBSimulatorControl)), I was able to get the tests to run and pass. 🚀

Since the UI tests need a different test-runner, I brought back the `ui_test` entry in `Makefile` to make a call to Buck with a custom config. The UI tests need a host app that is distinct from the target app. Instead of repurposing `ExampleApp` for it, I created an entirely separate host app with a dummy application delegate. It seems to do the job.

By repeatedly running the UI tests, I found out that:
 - the error messaging around them is extremely poor - for example, nearly any early runtime issue will be reported as an application launch timing out and require inspecting the Console logs to find that, for example, an dylib's image could not be loaded;
 - the tests fail periodically for seemingly no reason - the root of the failure is unclear, since Buck ends the log with a file handle error, but app installation errors like `Bundle Name (null) is not a String for CFBundleName` or `Application with bundle id 'com.airbnb.ExampleApp' is not installed` often preceed it;

Interestingly enough, the latter issue does not appear to bug [Facebook's example](https://github.com/facebook/buck/tree/master/test/com/facebook/buck/apple/testdata/apple_test_xcuitest). Perhaps BuckSample is heavy enough to periodically time out the application installer? For reference, this is the file handle error regularly appearing in the logs:

```2019-04-13 22:08:20.180-0700 New Error Built ==> Error Domain=com.facebook.FBControlCore Code=0 "File Handle is not open for reading, you should call 'startReading' first" UserInfo={NSLocalizedDescription=File Handle is not open for reading, you should call 'startReading' first}```

Either way, UI tests should now be buildable and runnable from CLI and able to leverage Buck caching. Hooray! 🎉

### CODE REVIEW

File: App/BUCK [hunk 1]
```diff
@@ -308,23 +311,53 @@ apple_test_lib(
...
 )
 
-# This test does not run from Buck CLI. However it can run from the generated Xcode project.
-# For more information, see https://github.com/airbnb/BuckSample/pull/70#issuecomment-481451036
+# This test suite cannot be packaged with other tests due to a dependency on `fbxctest` instead of `xctool`.
 apple_test_lib(
     name = "XCUITests",
     destination_specifier = {
-        "name": "iPhone 8",
+        "name": "iPhone X",
```
dfed: curious if this change is needed or just for fun 🙂

App/BUCK
```diff
@@ -308,23 +311,53 @@ apple_test_lib(
     ],
 )
 
-# This test does not run from Buck CLI. However it can run from the generated Xcode project.
-# For more information, see https://github.com/airbnb/BuckSample/pull/70#issuecomment-481451036
+# This test suite cannot be packaged with other tests due to a dependency on `fbxctest` instead of `xctool`.
```
dfed: I _love_ this comment. Thank you!

dfed: Curious: are these tests runnable from the generated Xcode project?

App/BUCK
```diff
@@ -308,23 +311,53 @@ apple_test_lib(
...
+        "name": "iPhone X",
     },
     run_test_separately = True,
-    test_host_app = ":ExampleApp",
+    test_host_app = ":XCUITestsHostApp",
     srcs = glob([
         "XCUITests/*.swift",
     ]),
     is_ui_test = True,
+    ui_test_target_app = ":ExampleApp",
```
dfed: Is this field documented anywhere? Would be great to have a comment here explaining what this line is doing

Makefile
```diff
@@ -46,6 +46,10 @@ test:
 	xcrun llvm-profdata merge -sparse "$(buck_out)/tmp/code-"*.profraw -o "$(buck_out)/gen/Coverage.profdata"
 	xcrun llvm-cov report "$(buck_out)/gen/App/ExampleAppBinary#iphonesimulator-x86_64" -instr-profile "$(buck_out)/gen/Coverage.profdata"
 
+fbxctest = tools/fbxctest/bin/fbxctest
+ui_test:
+	$(BUCK) test //App:XCUITests --config apple.xctool_path=$(fbxctest)
```
dfed: I would love a comment explaining why we’re setting a special tool path here

App/BuckSupportFiles/DummyAppDelegate.swift
```diff
@@ -0,0 +1,14 @@
+import UIKit
+
+#if swift(>=4.2)
+typealias ArgType = [UIApplication.LaunchOptionsKey: Any]? // Xcode 10
+#else
+typealias ArgType = [UIApplicationLaunchOptionsKey : Any]? // Xcode 9
```
dfed: I generally like writing framework code that supports multiple Xcode versions, but I don’t believe we have an intent of continuing support for Xcode 9 in this project.

qyang-nj: CI uses Xcode 9, and I couldn't get it work with Xcode 10 with Travis CI.

App/BUCK
qyang-nj: Do we need both `host_app` and `target_app` at the same time?

App/BUCK
bogo: I think I was getting some `invalid device` errors in Console.app without it. I just tested it right now though and it looks fine. Reverting to the iPhone 8!

One observation is that I was unable to add more specifics to `destination_specifier` entry – I think FBXCTool just uses the default iOS of the default Xcode on the system. Might be worth adding a comment, what do you think?

App/BUCK
bogo: @dfed This field is not officially documented, but it is mentioned in [the test fixture](https://github.com/facebook/buck/blob/master/test/com/facebook/buck/apple/testdata/apple_test_xcuitest/BUCK.fixture#L66). I'll add a comment about it.

@qyang-nj Unfortunately, yes. Removing either yields an instant `NO TESTS RAN` result. The test fixture I linked above has an `UITestLegacy` test library that doesn't have `ui_test_target_app`, but it doesn't seem to run. (or be run)

App/BuckSupportFiles/DummyAppDelegate.swift
dfed: 🤦‍♂️ @qyang-nj knows more than I do on this one

App/BUCK
dfed: > Might be worth adding a comment, what do you think?

Always worth adding a comment!

App/BuckSupportFiles/DummyAppDelegate.swift
bogo: Yep, this is mostly to pass tests on CI. Will keep it for now, sorry!

qyang-nj:
This is a very exciting PR! Thanks for the contribution.

Two more questions:
- Does this also work with generated Xcode project?
- Did you modify any source code of `fbxctest`?

App/BUCK
bogo: As-is, the tests are not runnable - they are not part of the `tests` field for `ExampleAppLibrary`. I will add them there to solve this issue.

dfed: Adding them under `tests` will probably break CI. Maybe we need to add the `ui_test` target to our [xcode_workspace_config](https://github.com/airbnb/BuckSample/blob/469a1fba5d084a08f2765dd5f67851289a2c9b43/App/BUCK#L122)?

bogo: And yes, after adding them there, the tests can be launched from Xcode! The host app is not used at all – not sure why it's necessary for CLI, but I am happy to kick that can down the road.

bogo:
@qyang-nj it works with the generated Xcode project as well! however, it does not use the host app and launched the target app instead. Which is pretty much what we want anyway. :)

`fbxctest` is copied verbatim from the Buck's test tooling. The only way I modified it is by adding `libShimulator.dylib` and `libMaculator.dylib` from [elsewhere in the FBSimulatorControl repo](https://github.com/facebook/FBSimulatorControl/blob/master/Shims/Binaries/libShimulator.dylib). I'll replace Facebook's README with a reference to FBSImulatorControl instead.

dfed:
@bogo please update the branch, and let me know when this is good to merge 🙂

App/BUCK
bogo: Actually, my initial assertion about `destination_specifier` might be untrue – despite spitting out `Current Runtime environment does not support Simulator Configuration` for incorrect combinations, setting up a correct combination of `name` and `OS` launches fine. (e.g. iOS 10.2 works with iPhone 6, but not iPhone 8)

OTOH, iOS 10.2 doesn't seem to be happy with `XCUITests`, so ¯\_(ツ)_/¯.

bogo: (in the end, comment might be unnecessary)

bogo:
Aight, I also moved the `ui_tests` to be `extra_tests` on the workspace level, since:
`> tools/buck test //App:ExampleAppLibrary`
Would actually fail, due to the unsupported `xctool` parameter.

While running "all" tests on the monolith makes sense, running tests on the workspace makes a bit less sense, so I'll keep them there for now. This is obviously not ideal in case if there are multiple applications in the workspace, but it should be fine for now.

I'd say we are ready for landing, as soon as we are green on Travis! 🚀

bogo:
```Test Suite XCUITests.xctest started
BuckSampleUITests
    ✓ testExampleAppLabel (5.177 seconds)
    ✓ testMainScreenLabelExists (4.745 seconds)
	 Executed 2 tests, with 0 failures (0 unexpected) in 9.922 (9.925) seconds```

🚀 at will, @dfed!

bachand:
This is awesome @bogo. Thanks for this contribution 🙏 One nit is that it would be nice to document the nuances of the different host apps (e.g. one is necessary for CLI but isn’t used for Xcode) in code comments. But not necessary for us to move forward. Thanks again 👍👍

gsabran:
This is awesome!
I've tried to run it without the Watch and iMessage extension. But simply removing the iMessage extension [here](https://github.com/airbnb/BuckSample/blob/master/App/BUCK#L150) makes the test fail, with no insightful logs.The dependency is not needed to build the Example app right?

bogo:
@gsabran, please create a separate issue with your report - FWIW, I think there's more to removing the iMessage/Apple Watch extensions.

PR #86, opened 2019-04-20
[WIP] Add example Siri Shortcut

bachand:


Config/buck_rule_macros.bzl
```diff
@@ -294,7 +296,63 @@ def mlmodel_resource(
...
+    genrule_name = "process_" + resource_dependency_name
+    # Create a genrule to process the .intentdefinition file. As far as we can tell, the
+    # `intentbuilderc` command makes no change to the .intentdefinition file. We nevertheless run it
+    # in an effort to be defensive.
+    logging_genrule(
+        name = genrule_name,
+        srcs = [definition_directory + definition_name + ".intentdefinition"],
+        bash = """
+        developer_dir=`xcode-select -p`
+        "$developer_dir/usr/bin/intentbuilderc" $SRCS $TMP ""
```
qyang-nj: It's easier to just call `xcrun intentbuilderc`.

bachand: Thanks @qyang-nj e08d22425328726122f14a50568a0a83b553ed80

App/BUCK
```diff
@@ -63,17 +63,27 @@ apple_library(
...
         "//Pods:CryptoSwift",
         "//Pods:PromiseKit",
         ":ExampleAppResource",
         ":ExampleAppAssets",
+        ":IntentsDefinition",
     ]
     + first_party_library_dependencies
     + build_phase_scripts,
+    frameworks = [
+        "$SDKROOT/System/Library/Frameworks/Intents.framework",
```
bachand: I'm not sure that `Intents.framework` is strictly necessary, but I have added it to reduce the potential causes of unexpected failure.

App/SiriShortcut/ExampleIntentUI/Base.lproj/MainInterface.storyboard
```diff
@@ -0,0 +1,30 @@
...
+        <deployment identifier="iOS"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="14490.49"/>
+        <capability name="Safe area layout guides" minToolsVersion="9.0"/>
+        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
+    </dependencies>
+    <scenes>
+        <!--Intent View Controller-->
+        <scene sceneID="7MM-of-jgj">
+            <objects>
+                <viewController id="ObA-dk-sSI" customClass="IntentViewController" customModule="ExampleIntentUIBinary" sceneMemberID="viewController">
```
bachand: `customModule` must be explicitly set in order to get Buck CLI to work. This is noted in a comment in `App/Buck`.

App/SiriShortcut/_IntentsCompiler/IntentsCompiler.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,329 @@
+// !$*UTF8*$!
```
bachand: This Xcode project exists for running code generation on the `.intentdefinition`. The `README.md` explains more.

App/BUCK
bachand: Nevermind. This is used by the generated intent source code.

bachand: So it's good we keep `Intents.framework` 👍

App/BUCK
```diff
@@ -159,6 +172,128 @@ apple_package(
...
+    deps = [
+        ":IntentsDefinition",
+    ],
+    frameworks = [
+        "$SDKROOT/System/Library/Frameworks/Intents.framework",
+    ],
+)
+
+apple_bundle(
+    name = intent_bundle_name,
```
bachand: If necessary you can also set `product_name` to name the `.appex` file and its associated binary, but that's only necessary if you want `name` to be different.

bachand: By default, `product_name == name`

Config/buck_rule_macros.bzl
```diff
@@ -294,7 +296,103 @@ def mlmodel_resource(
...
+        intent_interface="`find "$TMP" -name %sIntent.swift`"
+
+        if [[ -z "$intent_interface" ]]; then
+            echo "Compiler xcodeproj produced no Swift interface for the provided intent"
+            echo "Are you sure the .intentdefinition defines the intent?"
+            exit 1
+        fi
+    else
+        intent_interface="$TMP/Dummy.swift"
+        echo "class Dummy { }" > "$intent_interface"
```
bachand: I think I should update this to copy a pre-built version of `BuckPhotoIntent.swift`. If I do that I think the PR will build.

App/AppDelegate.swift
```diff
@@ -13,10 +13,46 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+  func application(
+    _ application: UIApplication,
+    continue userActivity: NSUserActivity,
+    restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool
+  {
+    return continueWithUserActivity(userActivity)
+  }
+  #else
+  // Xcode 9
+  func application(
```
bachand: Can't wait to upgrade CI to get rid of this stuff.

App/SiriShortcut/_Placholder_BuckPhotoIntent.swift
```diff
@@ -0,0 +1,89 @@
+
+//
+// BuckPhotoIntent.swift
+//
+// This file was automatically generated and should not be edited.
+//
+
+import Foundation
```
bachand: Temporary until we upgrade CI.

.buckconfig
```diff
@@ -8,7 +8,7 @@
 
 [swift]
   version = 4.0
-  compiler_flags = -DBUCK -enable-testing -g -Onone -whole-module-optimization $(config custom.other_swift_compiler_flags)
+  compiler_flags = -DBUCK -enable-testing -g -Onone -whole-module-optimization $(config custom.other_swift_compiler_flags) -DDISABLE_SIRI_SHORTCUT
```
bachand: It's behind a feature flag.

bachand: This works around having Xcode 9 in CI.

bachand:
Almost done...

are broken with this error

```
    Falling back to contents of entitlements file "XCUITestsHostAppBinary.entitlements" because it was modified during the build process. Modifying the entitlements file during the build is unsupported.error: The file “XCUITestsHostAppBinary.entitlements” couldn’t be opened because there is no such file.`
```

I may need to map different entitlements to different binaries.

Config/buck_rule_macros.bzl
```diff
@@ -294,7 +296,113 @@ def mlmodel_resource(
...
+    fi
+
+    cp "$intent_interface" "$OUT"
+    """ % intent_name
+
+    logging_genrule(
+        name = interface_source_name,
+        srcs = [
+            compiler_xcodeproj,
```
bachand: TODO: we'll need to pass the .intentdefinition here too in order to invalidate the rule when we make changes to that file.

App/BUCK
```diff
@@ -158,6 +169,129 @@ apple_package(
...
+    srcs = glob([
+        "SiriShortcut/ExampleIntent/**/*.swift",
+    ]) + [":BuckPhotoIntentInterface"],
+    configs = intent_binary_configs(intent_info_plist_substitutions),
+
+    # "-e _NSExtensionMain" tell linker this binary is app extension, so it won't fail due to
+    # missing _main "-Xlinker -rpath -Xlinker @executable_path/../../Frameworks" tells the
+    # executable binary to look for frameworks in ExampleApp.app/Frameworks instead of PlugIns, so
+    # that we don't need to have the libSwift*.dylib in ExampleApp.app/PlugIns/*.appex/Frameworks
+    linker_flags = [
```
rockbruno: I think you need to add `/usr/lib/swift` as a search path to make it work on iOS 12.2 forward. It was the case for me:

```
    linker_flags = [
        "-e",
        "_NSExtensionMain",
        "-Xlinker",
        "-rpath",
        "-Xlinker",
        "/usr/lib/swift"
        "-Xlinker",
        "-rpath",
        "-Xlinker",
        "@executable_path/../../Frameworks",
    ],
```

bachand: Interesting. Thanks for sharing @rockbruno. Are you on Swift 5?

bachand: I’m trying to understand if this is because the Swift libraries packaged with the app are now being ignored. If I understand correctly this occurs with Swift 5+ and iOS 12.2+.

rockbruno: I'm on 4.2, but I'm using Xcode 10.2 which IIRC uses the Swift 5 compiler regardless of the version.
I remember they changed Buck to add this import by default so I'm surprised I had to do this, so maybe it's just the extension bundles that weren't covered by the change...

PR #90, opened 2019-04-27
Fix Xcode project and `make debug`

bachand:
- https://github.com/airbnb/BuckSample/pull/87 broke building and running with the Xcode project. `info_plist_substitutions` will work for Buck CLI builds, but these substitutions need to be added to `configs` for Xcode builds to work. cc: @gsabran 
- Remove `Phone :` from `make debug` command that was preventing the simulator from launching.

The app crashes on launch with `make debug`, and I've tracked that issue here: https://github.com/airbnb/BuckSample/issues/91

### DIFF
--- a/App/BUCK
+++ b/App/BUCK
@@ -166,6 +166,7 @@ apple_package(
 # binary, the Copy Files phase to embed the watch app and extension will search the `iphoneos` directory,
 # and fail the build on generated Xcode projects.
 
+# This is the code that runs on the iPhone and talks to the app running on the Watch.
 apple_binary(
     name = "ExampleWatchAppExtensionBinary",
     srcs = glob([
@@ -203,6 +204,12 @@ apple_bundle(
     xcode_product_type = "com.apple.product-type.watchkit2-extension",
 )
 
+# This is the code that runs on the Watch.
+apple_binary(
+    name = "ExampleWatchAppBinary",
+    configs = watch_binary_configs("ExampleApp.WatchApp")
+)
+
 apple_bundle(
     name = "ExampleWatchApp",
     binary = ":ExampleWatchAppBinary",
@@ -223,11 +230,6 @@ apple_bundle(
         ":ExampleWatchAppExtension",
         ":ExampleWatchAppResources",
     ],
-)
-
-apple_binary(
-    name = "ExampleWatchAppBinary",
-    configs = watch_binary_configs("ExampleApp.WatchApp")
 )
 
 apple_resource(
--- a/App/WatchApplication/Info.plist
+++ b/App/WatchApplication/Info.plist
@@ -18,7 +18,7 @@
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
 	<string>1.0</string>
-		<key>CFBundleSignature</key>
+	<key>CFBundleSignature</key>
 	<string>????</string>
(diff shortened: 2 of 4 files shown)


App/BUCK
```diff
@@ -203,6 +204,12 @@ apple_bundle(
     xcode_product_type = "com.apple.product-type.watchkit2-extension",
 )
 
+# This is the code that runs on the Watch.
```
bachand: I moved the `apple_binary` before the `apple_bundle` for consistency with `ExampleWatchAppExtension`.

Makefile
```diff
@@ -25,7 +25,7 @@ message:
 	$(BUCK) build //App:ExampleMessageExtension
 
 debug:
-	$(BUCK) install //App:ExampleApp --run --simulator-name 'Phone: iPhone XS'
+	$(BUCK) install //App:ExampleApp --run --simulator-name 'iPhone XS'
```
bachand: @qyang-nj made this fix for the Airbnb app. His guess was that this was required in a previous version of the tooling.

App/WatchApplication/Info.plist
```diff
@@ -18,7 +18,7 @@
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
 	<string>1.0</string>
-		<key>CFBundleSignature</key>
```
dfed: 👏

Config/configs.bzl
```diff
@@ -87,15 +87,19 @@ def info_plist_substitutions(name):
     }
     return substitutions
 
+# TODO: Right now this is being used for both the code that runs on the Watch and the extension
+# that runs on the iPhone. I wonder if it would make sense to break this into two methods.
```
dfed: Instead of a TODO to break this up, maybe a comment that “If the build configuration for the watch extension needs to differ from the build configuration for the watch app, we should create a new macro.”

My thinking is that right now we don’t have a use case that we know of where we’d need to execute the TODO, so a TODO seems a little misleading (it says we want to do something, when instead we’re reminding ourselves that we might need to do something). Leaving a comment showing that we didn’t have explicit intent for the two binaries to have the same configuration, however, does feel helpful.

Edit: after writing the above I’ve now noticed that we’re setting two bundle identifiers in the below method. Maybe the TODO could state that we should split this macro up due to the fact that the configurations are already diverging?

Makefile
```diff
@@ -35,7 +35,7 @@ ci: install_buck targets build test ui_test project xcode_tests watch message
 
 
 buck_out = $(shell $(BUCK) root)/buck-out
-test:
```
dfed: 👏

did you use a tool to format or did you happen to find these white space issues as you were perusing?

bachand: I've set up VS Code to automatically trim trailing whitespace. So whenever I make an edit to a file, it will clean things up.

https://github.com/bachand/battlestation/blob/8a4ce7bf9f0995cde713981ec0471e2b52f4e746/config/vscode_settings.json#L5

Config/configs.bzl
bachand: Good idea. I improved the comment: https://github.com/airbnb/BuckSample/pull/90/commits/c06e7fe63cb00976eb5e5d71b804a0835d3b5cb7

PR #92, opened 2019-04-27
Create pull_request_template.md

bachand:
It seems ideal that we ask people to build and run to avoid having the project break.

cc: @qyang-nj @dfed @shepting @fdiaz


bachand:
Good idea.

I have added four issues. If we could get all four of these working that would be huge!

Issue #94, opened 2019-04-28
Update CI to verify that the app successfully runs when built via Buck CLI


bachand:
This issue is a follow-up to https://github.com/airbnb/BuckSample/pull/92. It takes https://github.com/airbnb/BuckSample/issues/93 one step further by verifying that the app actually _runs_ in CI. This can be thought of as a launch test.

qyang-nj:
@bachand I forgot what this was about. I think we are running `make ci` on CI which includes `buck build`.

PR #99, opened 2019-04-28
[Blocked on CI upgrade] Test that app installs to simulator via CLI in CI

### LINKED ISSUE #93: Update CI to verify that the app successfully installs via Buck CLI (`make debug`)
This issue is a follow-up to https://github.com/airbnb/BuckSample/pull/92 to verify the `make debug` successfully installs the app to the simulator.

bachand:
Fixes #93. We cannot remove "Built and ran with Buck CLI (`make debug`)" from the PR template until #94 is fixed as well. And we can't do that until we get the app to actually run via the CLI again 😬

I verified that when `make debug` times out it fails with a non-zero exit code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ debug:
 targets:
 	$(BUCK) targets //...
 
-ci: install_buck targets build test ui_test project xcode_tests watch message
+ci: install_buck targets build test ui_test project xcode_tests watch message debug
 	echo "Done"
 
 
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ message:
 	$(BUCK) build //App:ExampleMessageExtension
 
 debug:
-	$(BUCK) install //App:ExampleApp --run --simulator-name 'iPhone XS'
+	$(BUCK) install //App:ExampleApp --run --simulator-name 'iPhone X'
 
 targets:
 	$(BUCK) targets //...


dfed:
Almost there! Build failure is:
```
BUILD FAILED: Cannot install //App:ExampleApp#dwarf-and-dsym,no-include-frameworks (no appropriate simulator found)
```

We might want to change our `Makefile` so that `debug` that doesn’t specify a specific simulator. Alternatively, we may be able to get away with an iPhone X, SE or another device that is known to Xcode 9 (currently we’re selecting an XS).

@bachand or @qyang-nj: do you remember why we are specifying a specific sim [in our Makefile](https://github.com/airbnb/BuckSample/blob/e5e2da733d4a2a0e87e22fe85d147946d9d9adff/Makefile#L28)?

bachand:
I don't know why we added this. I think it's been around awhile. Probably added it so we didn't end up with something random.

I tried iPhone X. No luck.

```
BUILD FAILED: Cannot install //App:ExampleApp#dwarf-and-dsym,no-include-frameworks (could not start simulator iPhone X within 20000 ms)
INSTALLING: FINISHED IN 1m 29.0s
```

qyang-nj:
If we don't give a name, it's iPhone 5s by default. I guess we wanted a high-resolution simulator.

qyang-nj:
Probably we need `Phone: iPhone X` for Xcode 9. That's why we added there. Just a guess. I don't have Xcode 9 any more.

bachand:
Ok let’s put this on ice until we upgrade CI

Issue #97, opened 2019-04-28
Unable to run `ExampleWatchApp` from Buck-generated Xcode project


bachand:
### Steps
1. `make project`.
2. Create a scheme for `ExampleWatchApp#watchsimulator-i386`.
3. Build and run.

### Result
Both the iPhone and Watch simulators open. The app is not installed on the watch. 

### Other attempts

- I tried explicitly installing `ExampleWatchApp` via the Watch app on the iPhone simulator.
- I tried creating a scheme for `ExampleWatchAppExtension#watchsimulator-i386` instead and then building and running.

Neither of these worked.

bachand:
cc: @qyang-nj who has the most knowledge about the watch.

qyang-nj:
Fixed by #100

PR #100, opened 2019-04-29
Fix watch launch problem in generated project

qyang-nj:

Fix issue #97.

@bachand

### DIFF
--- a/App/WatchApplication/Info.plist
+++ b/App/WatchApplication/Info.plist
@@ -18,18 +18,16 @@
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
 	<string>1.0</string>
-	<key>CFBundleSignature</key>
-	<string>????</string>
 	<key>CFBundleSupportedPlatforms</key>
 	<array>
 		<string>WatchSimulator</string>
 	</array>
 	<key>CFBundleVersion</key>
 	<string>1</string>
-	<key>DTPlatformName</key>
-	<string>watchsimulator</string>
-	<key>DTSDKName</key>
-	<string>watchsimulator2.0</string>
+	<key>UIDeviceFamily</key>
+	<array>
+		<string>4</string>
+	</array>
 	<key>UISupportedInterfaceOrientations</key>
 	<array>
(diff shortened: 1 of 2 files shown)


App/WatchExtension/Info.plist
```diff
@@ -32,9 +32,5 @@
 	</dict>
 	<key>WKExtensionDelegateClassName</key>
 	<string>$(PRODUCT_MODULE_NAME).ExtensionDelegate</string>
-	<key>UIRequiredDeviceCapabilities</key>
-	<array>
-		<string>arm64</string>
```
qyang-nj: This is the culprit. Others are clean-up.

App/WatchApplication/Info.plist
```diff
@@ -18,18 +18,16 @@
...
 	<key>CFBundleSupportedPlatforms</key>
 	<array>
 		<string>WatchSimulator</string>
 	</array>
 	<key>CFBundleVersion</key>
 	<string>1</string>
-	<key>DTPlatformName</key>
-	<string>watchsimulator</string>
-	<key>DTSDKName</key>
-	<string>watchsimulator2.0</string>
```
bachand: Why are these not necessary?

App/WatchExtension/Info.plist
bachand: Huh. Since this is running on the phone it’s confusing why it would break...

qyang-nj: Simulator isn't arm64.

App/WatchApplication/Info.plist
qyang-nj: Based on my limit observation, DT prefix indicates they are internal Xcode settings, which are usually inserted by Xcode automatically. Unlike other keys, those DT keys don't have any official doc to explain them.

App/WatchExtension/Info.plist
bachand: Got it. I recall we had to add this somewhere in order to upload the App Store. Hopefully it was just necessary for the main app and not this extension.

Issue #101, opened 2019-04-29
Siri does not read or display response text for Siri Shortcut


bachand:
This issue is related to https://github.com/airbnb/BuckSample/pull/86 (which is unmerged due to https://github.com/airbnb/BuckSample/issues/102).

That PR gets a Siri Shortcut working with `ExampleApp`. Siri displays the custom Siri Shortcut UI but fails to display or read out the response text.

The response text for the `.success` case should be `Here is a wild buck.`

As I wrote https://github.com/airbnb/BuckSample/pull/86, I built the same example Siri Shortcut in a manually managed Xcode project. In that project, the response text is displayed and read out as you would expect. I have tried diff'ing the resulting `.app` files and can't the culprit. The code for the manually managed example project is here: [XcodeExampleApp.zip](https://github.com/airbnb/BuckSample/files/3126037/XcodeExampleApp.zip).

Here are the `.app` files and dSYMs for the working manually managed Xcode project and the broken Buck-generated project: [01 ExampleApp - correct.zip](https://github.com/airbnb/BuckSample/files/3126054/01.ExampleApp.-.correct.zip) and [02 ExampleApp - no Siri phrase.zip](https://github.com/airbnb/BuckSample/files/3126055/02.ExampleApp.-.no.Siri.phrase.zip).

| Response text successfully read out in manually managed Xcode project | No response text in Buck-generated Xcode project | 
| ---- | ---- |
|  ![Recording](https://user-images.githubusercontent.com/1791049/56877917-2d11cc80-6a06-11e9-9797-d9dd44a90c7a.gif) | ![Recording](https://user-images.githubusercontent.com/1791049/56877921-36029e00-6a06-11e9-8a2f-1abdae1acddf.gif) |

I have looked through the console logs of both the working and not working apps using Console.app (filtering for the term "example"). Sometimes (but not always) the broken Buck-built app will print out errors like this:

```
error	13:16:13.862344 -0700	assistant_service	-[AAPIntentsInfoSyncHandler getChangeAfterAnchor:changeInfo:] com.apple.siri.applications: No intents are supported by Intents extensions for com.airbnb.ExampleApp: (
    "<LSPlugInKitProxy: 0x7fbb55c4b160> pluginID=com.airbnb.ExampleApp.IntentExtension UUID=FFC456C5-061C-4B6E-B629-26DDA91FF66E version=1.0"
)
```

[...]

PR #103, opened 2019-05-01
Consolidate on `en` for development language

bachand:
I believe this is the root cause problem of https://github.com/airbnb/BuckSample/issues/101.

Right now we are setting this to `en-us` in some places and `Swift` in other 🤦‍♂ 

In a standard Xcode project, this defaults to `en`. Apple suggested I set it to `en`.

In our production app, we are inconsistent. In one target we have it as `English` (which is invalid). In two others I checked we have it as `en`.

It feels inelegant setting a global variable, but this also feels like a step in the right direction of being more DRY.

Thank you Apple for your help, if you're watching 😄 

I'm leaving that issue open until I rebase my other branch.

cc: @qyang-nj @dfed

### DIFF
--- a/Config/configs.bzl
+++ b/Config/configs.bzl
@@ -1,5 +1,7 @@
 load("//Config:utils.bzl", "config_with_updated_linker_flags", "configs_with_config")
 
+DEVELOPMENT_LANGUAGE = "en"
+
 def pretty(dict, current = ""):
     current = "\n"
     indent = 0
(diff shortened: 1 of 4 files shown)


Config/configs.bzl
```diff
@@ -53,7 +55,7 @@ def library_configs():
 def app_binary_configs(name):
     binary_specific_config = {
         "ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES": "YES",
-        "DEVELOPMENT_LANGUAGE": "Swift",
```
qyang-nj: 😱😱😱

PR #104, opened 2019-05-01
homepage]: https://www.contributor-covenant.org

koolaid28:
ayehomepage]: https://www.contributor-covenant.org

- [ ] Built and ran in Xcode (`make project`)


bachand:
@ljharb can you advise on whether we should add this code of conduct?

I am not sure what our strategy is for that across Airbnb repos.

bachand:
Thanks for the PR @koolaid28. We’re not ready to make a change of this nature, but I fully agree that we should conduct ourselves in this repo in a way that makes everyone feel like they belong.

PR #111, opened 2019-06-05
WIP - buck caching

thedavidharris:
- [ ] Built and ran in Xcode (`make project`)

Need to figure out what the CIRCLE_BRANCH_NAME variable is on forks to test this... but structure I think would be something like this.

There's a cleaner way to do this in configs as well where you extract things into `commands` that you use as building blocks for `jobs` and then orchestrate them differently on branches, but this is a start

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,6 @@ jobs:
       TERM: dumb
     steps:
       - checkout
-      
       - restore_cache:
           name: Restore Buck build cache
           keys: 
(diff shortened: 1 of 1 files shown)


qyang-nj:
@thedavidharris just turned it on

qyang-nj:
@thedavidharris is this still WIP?

thedavidharris:
Yeah, it’s never actually been working and not quite sure why, haven’t looked at it in a bit. Will take it again.

PR #114, opened 2019-06-20
Add some helpful commands to README

bachand:
I think this makes the project more accessible.

cc: @qyang-nj @dfed @shepting 


bachand:
Oh, and @fdiaz

bachand:
Good idea @qyang-nj . I updated it with a description that I workshopped with @dfed and @fdiaz

PR #116, opened 2019-07-14
Ignore 3rd party dependencies when computing coverage

rockbruno:
Hi there! I excluded the Pods and the generated CoreML file from being included in the coverage report so we can get a more realistic view of the app's code coverage.


qyang-nj:
LGTM, Thanks!

PR #122, opened 2019-09-25
Move AFNetworking to binary cocoapod

thedavidharris:
might not want to vendor the pods folder in this approach

- [x ] Built and ran in Xcode (`make project`)
- [x ] Built and ran with Buck CLI (`make debug`)

Tests didn't seem to be working on my machine, everything runs fine, was concerned about tests mostly for the embedding.


thedavidharris:
Just https://github.com/airbnb/BuckSample/blob/0a4a4d2919b05ba74da4791def260174556a06c0/Libraries/SwiftWithPrecompiledDependency/Sources/SwiftWithPrecompiledDependencyClass.swift, if there's probably an easier lib that could get used without dipping into networking to check things are working

PR #134, opened 2020-07-02
Open source Buck Local 

xianwen:
Buck Local is an internal tool developed in Airbnb.
It generates an Xcode project that invokes Buck to build the App when you hit the Build button in Xcode. This makes local iOS builds 2x faster without Buck Http cache, and iOS complete builds 5~6 times faster when Buck Http cache is enabled.

To apply Buck Local in your own iOS project, you can do so by:

1. Copy `Rakefile` file and `BuckLocal` dir from BuckSample to the root directory of your own iOS project
2. Replace all `apple_asset_catalog` with `buck_local_apple_asset_catalog`, and `apple_resource` with `buck_local_apple_resource` in your BUCK files.
3. Migrate the top level `apple_binary` in your BUCK file to `buck_local_binary`, `xcode_workspace_config` to `buck_local_workspace`, and `apple_bundle` to `buck_local_bundle` as demonstrated in this PR.

After these steps, you'll be able to generate your own Buck Local project by calling `rake buck_local:generate_project`. You'll be able to build / run / debug your app in Xcode as if it's a normal Xcode project, while the local builds are more than 2 times faster.

- [X] Built and ran in Xcode (`make project`)


BuckLocal/BUCK
```diff
@@ -0,0 +1,24 @@
+# Build all libraries via Buck, and copy over the artifacts to DerivedData so they can be used by Xcode.
+xcode_prebuild_script(
+    name = "BuckLocalBuildAndCopy",
+    cmd = "$REPO_ROOT/BuckLocal/buck_build_and_copy.sh"
+)
+
+# This is the target that the `buck_local_binary` target depends on, in order to utilize Buck for building all modules.
+apple_library(
+    name = "BuckLocal",
```
qyang-nj, line 9: Isn't this file generated and shouldn't be checked in?

App/BUCK
```diff
@@ -120,11 +124,11 @@ apple_test_all(
     prebuilt_frameworks = prebuilt_frameworks,
 )
 
-xcode_workspace_config(
+buck_local_workspace(
     name = "workspace",
     workspace_name = "ExampleApp",
     src_target = ":ExampleApp",
-    additional_scheme_actions = {
+    native_xcode_scheme_actions={
```
bachand, line 131: Much better name 👌

App/BUCK
```diff
@@ -145,12 +149,12 @@ apple_bundle(
...
     info_plist_substitutions = info_plist_substitutions("ExampleApp"),
-    deps = [
+    native_xcode_deps=prebuilt_frameworks + [
         # For "#watch", https://buckbuild.com/rule/apple_bundle.html#deps
         ":ExampleWatchApp#watch",
         ":ExampleMessageExtension"
-    ]
-    + prebuilt_frameworks,
+    ],
+    buck_local_deps=prebuilt_frameworks,
```
bachand, line 157: I hope we can get watch and iMessage working some day.

BuckLocal/BUCK.erb
```diff
@@ -0,0 +1,24 @@
+# Build all libraries via Buck, and copy over the artifacts to DerivedData so they can be used by Xcode.
+xcode_prebuild_script(
```
bachand, line 2: Can you help me understand why this is a pre-build step? I'd expect it to be a post-build step.

BuckLocal/RemoteSourcePathList.txt
```diff
@@ -0,0 +1 @@
+.
```
bachand, line 1: Do we need this?

BuckLocal/SourcePathRemapping_Postfix
```diff
@@ -0,0 +1,3 @@
+</dict>
+</dict>
+</plist>
```
bachand, line 3: Do we need this?

BuckLocal/buck_build_and_copy.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
bachand, line 1: Sometimes Bash is the best tool for the job.

BuckLocal/ruby_scripts/builder.rb
```diff
@@ -0,0 +1,143 @@
...
+# require 'tooling/buck_cache_parser'
+# require 'tooling/path'
+# require 'tooling/files'
+
+require 'buck_log_formatter'
+require 'common'
+require 'json'
+require 'fileutils'
+
+module BuckLocal
```
bachand, line 11: I think one day this could grown into a Swift command line tool

BuckLocal/ruby_scripts/common.rb
```diff
@@ -0,0 +1,266 @@
+require 'English'
+require 'open3'
+require 'fileutils'
+
+# Wrap text in terminal color codes
+module Color
```
bachand, line 6: It could be cool to put this code into a Ruby get that we host on public Github so we can share the code between our public and private repos.

bachand:
(My comments are not blocking)

BuckLocal/BUCK
xianwen: @qyang-nj , I haven't been able to get `rule_exists` (https://buck.build/skylark/generated/rule_exists.html) working properly to capture whether BUCK file has been generated or not.

Currently, in Airbnb, we're still using Python3, so we have been using `glob` to get the job done. However, `glob` is not available in Skylark, and https://buck.build/skylark/generated/rule_exists.html is supposed to be a replacement for it.

Until we can find a way to get `rule_exists`, we unfortunately still needs to check in this BUCK file so that it can be picked up by the generation process.

App/BUCK
xianwen: I'm sure we can get them working once its priority is high enough for us to plan for it.

BuckLocal/BUCK.erb
xianwen: Before we ask Xcode to build the app, we need to build all the libs and copy them to derived data so that they could be found by Xcode. Then, we let Xcode start building the app, which is mostly linking the libs and packaging the App bundle.

BuckLocal/RemoteSourcePathList.txt
xianwen: Yes, it's used in `remap_debug_source_path.sh`

BuckLocal/SourcePathRemapping_Postfix
xianwen: Yes, it's used in `remap_debug_source_path.sh`

BuckLocal/buck_build_and_copy.sh
xianwen: :)

BuckLocal/ruby_scripts/builder.rb
xianwen: Yep, that'd be nice.

App/BUCK
bachand: Yes, not a priority today 👍

BuckLocal/ruby_scripts/common.rb
xianwen: Yes, that would definitely be nice.

xianwen:
> Let's update the README. Also tests 🥺

@bachand, good idea, I'll update README accordingly.

Actually, we do have lots of tests for Buck Local in our internal repo. I simply haven't migrated them yet.
Thanks for the reminder, I'm going to update this PR to migrate those unit tests as well.

bachand:
Awesome @xianwen. Thanks for addressing my feedback and, more importantly, for making this technology available to the community. I think that a lot of teams could benefit from this work, and the ideas behind it.

PR #137, opened 2020-08-17
Add new updated list of all ruletypes.

shepting:
I've been able to figure out how to generate this list again. It has been immensely helpful in the past to see _all_ of the potential arguments to various rules, even if they are undocumented to get an idea of what might be possible (and then to read source/tests from there).

## Tested by
- [x] Docs only change, did not test at all.
- [ ] ~Built and ran in Xcode (`make project`)~
- [ ] ~Built and ran with Buck CLI (`make debug`)~

## Please review
@xianwen @qyang-nj @bachand @dfed


Docs/all_ruletypes.py
```diff
@@ -0,0 +1,2893 @@
...
+    app_icon = None,
+    compatible_with = None,
+    default_target_platform = None,
+    dirs = None,
+    labels = None,
+    launch_image = None,
+    licenses = None,
+):
+    ...
+def apple_binary (
```
shepting, line 435: These `apple_xxxx` rules are particularly interesting for us.

Docs/all_ruletypes.py
```diff
@@ -0,0 +1,2893 @@
+# Generated by Steven Hepting on August 14th, 2020:
+#    buck audit ruletypes > ruletypes.txt
+#    cat ruletypes.txt | xargs -I _ buck audit ruletype _ > all_ruletypes.py
```
dfed, line 3: Love that you've documented how you created this script. However, I think running this command will end up deleting this documentation. Maybe we move this command documentation to a separate file or script?

qyang-nj: Did you generate this from `facebook/buck` master branch or dev, or `airbnb/buck`? It's better to comment the buck commit SHA as well.

shepting: I generated from our local version of Buck `buck version 23d5c09ab5491bd03fab3bf60cf15f6f87547fd6`

Docs/generate_rule_interface_listing.sh
```diff
@@ -0,0 +1,25 @@
...
+
+RULETYPE_FILE=all_ruletypes.py
+BUCK=$(command -v buck)
+
+echo "Generating all ruletype interfaces from local version of Buck: ${BUCK}"
+
+echo "# Generated by $(id -F)" > "${RULETYPE_FILE}"
+echo "# To re-generate run:" > "${RULETYPE_FILE}"
+echo "#    ./generate_ruletype_docs.sh" > "${RULETYPE_FILE}"
+echo "# Buck version: $($BUCK --version)" > "${RULETYPE_FILE}"
```
dfed, line 17:
```suggestion
echo "# Buck version: $($BUCK --version)" >> "${RULETYPE_FILE}"
```

Docs/generate_rule_interface_listing.sh
```diff
@@ -0,0 +1,25 @@
...
+cd "$(dirname "$0")" || exit # Run from the directory of this script
+
+RULETYPE_FILE=all_ruletypes.py
+BUCK=$(command -v buck)
+
+echo "Generating all ruletype interfaces from local version of Buck: ${BUCK}"
+
+echo "# Generated by $(id -F)" > "${RULETYPE_FILE}"
+echo "# To re-generate run:" > "${RULETYPE_FILE}"
+echo "#    ./generate_ruletype_docs.sh" > "${RULETYPE_FILE}"
```
dfed, line 16:
```suggestion
echo "#    ./generate_ruletype_docs.sh" >> "${RULETYPE_FILE}"
```

Docs/generate_rule_interface_listing.sh
```diff
@@ -0,0 +1,25 @@
...
+
+cd "$(dirname "$0")" || exit # Run from the directory of this script
+
+RULETYPE_FILE=all_ruletypes.py
+BUCK=$(command -v buck)
+
+echo "Generating all ruletype interfaces from local version of Buck: ${BUCK}"
+
+echo "# Generated by $(id -F)" > "${RULETYPE_FILE}"
+echo "# To re-generate run:" > "${RULETYPE_FILE}"
```
dfed, line 15:
```suggestion
echo "# To re-generate run:" >> "${RULETYPE_FILE}"
```

Docs/generate_rule_interface_listing.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -o errexit # fail if any command fails
+set -o pipefail
+set -o nounset
+
+cd "$(dirname "$0")" || exit # Run from the directory of this script
```
dfed, line 7: curious why we need the `|| exit`? I would expect the `set -o errexit` to do that for us.

shepting: Yeah, it probably should. I think this was a fix suggested by `shellcheck` for a _different_ script and I ended up copying it.

Docs/generate_rule_interface_listing.sh
shepting: 😆  I ended up fixing this locally before seeing your comment. Great catch.

PR #141, opened 2020-12-05
Add graph generation to Makefile

thexande:
One great use case for this repo is to prototype your project's dependency DAG with Buck. Instead of generating a project, I found it was much easier to arrange frameworks by generating a chart to reflect the current Buck DAG. 

I'm not sure if others will find it helpful, but I did as I worked to migrate our ~25 internal and ~35 external framework project into the Buck build system.

make_dag


Makefile
```diff
@@ -95,3 +95,5 @@ project: clean
 buck_local_project: clean
 	bundle exec rake buck_local:generate_project buck_binary_path=$(BUCK) workspace_target='//App:workspace-buck-local' top_level_lib_target='//App:ExampleAppLibrary' xcworkspace='App/ExampleAppBuckLocal.xcworkspace'
 	open App/ExampleAppBuckLocal.xcworkspace
+graph:
```
qyang-nj, line 98: Can we call this `dependency_graph` which is clearer?

qyang-nj (approved):
Thanks for making this PR. We have this command in the README, but it's nice to be a make task.

PR #142, opened 2020-12-07
Upgrade to Xcode 12 and fix all issues

xianwen:
1. Use new Buck version that supports Big Sur
2. Switch to the same `airbnb-ios` version of Buck, and add the `-BUCK` suffixes to makefile so that all jobs are still working.
3. Use new CirceCI VM that runs Xcode 12.2.0
4. Upgrade XCUITestRunner binary to be compatible with Xcode 12
5. Fix Xcode 12 compile errors

With this PR, we get `make ui_test` working again with Xcode 12 on Big Sur.

However, there are **two jobs stopped working** after this migration to Xcode 12: **UnitTestsWithHostApp, and Code Coverage.**
I think we can fix these 2 jobs in following up PRs, since otherwise nothing is working right now under Xcode 12.

@qyang-nj @shepting


.circleci/config.yml
```diff
@@ -2,14 +2,15 @@ version: 2
 jobs:
   build-and-test:
     macos:
-      xcode: "11.3.1"
+      xcode: "12.2.0"
```
shepting, line 5: :100:

Makefile
```diff
@@ -8,7 +8,7 @@ log:
 	echo "Make"
 
 install_buck:
-	curl https://jitpack.io/com/github/airbnb/buck/4bd1a08625454c5034eb6ef6193e94f9e6e62a62/buck-4bd1a08625454c5034eb6ef6193e94f9e6e62a62.pex --output tools/buck
+	curl https://jitpack.io/com/github/airbnb/buck/41a475f76f8c9223840af7054cbb8c2a2aa55192/buck-41a475f76f8c9223840af7054cbb8c2a2aa55192-java11.pex --output tools/buck
```
shepting, line 11: 👍

shepting (approved):

qyang-nj (approved):
Thank you!

PR #143, opened 2020-12-16
Bring back code coverage

qyang-nj:
Bring back code coverage and use the solution close to our internal implementation.

Please review: @xianwen @shepting


xianwen (approved):
Thank you!

PR #145, opened 2020-12-17
Fix BuckLocal compilation on Xcode 12 and Big Sur

xianwen:
image

@qyang-nj @shepting 

### DIFF
--- a/BuckLocal/remap_debug_source_path.sh
+++ b/BuckLocal/remap_debug_source_path.sh
@@ -15,6 +15,8 @@
 #
 DSYM_BUNDLE_PATH=$DWARF_DSYM_FOLDER_PATH/$DWARF_DSYM_FILE_NAME
 
+# Wait for 1 second to make sure the DSYM bundle had been generated.
+sleep 1
 # Use dwarfdump to get the UUID of the DWARF file in the dSym bundle.
 # The filename of the pList must match this UUID for the source path remapping to work.
 UUID=`dwarfdump --uuid $DSYM_BUNDLE_PATH | cut -d ' ' -f 2`
--- a/BuckLocal/ruby_scripts/project_generator.rb
+++ b/BuckLocal/ruby_scripts/project_generator.rb
@@ -89,7 +89,7 @@ module BuckLocal
     # Disable `build_implicit_dependencies` and `parallelize_buildables` for "BuckLocal" xcscheme.
     #
     def update_main_scheme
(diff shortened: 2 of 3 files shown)


BuckLocal/remap_debug_source_path.sh
```diff
@@ -15,6 +15,8 @@
 #
 DSYM_BUNDLE_PATH=$DWARF_DSYM_FOLDER_PATH/$DWARF_DSYM_FILE_NAME
 
+# Wait for 1 second to make sure the DSYM bundle had been generated.
+sleep 1
```
shepting, line 19: Oh man, I'm glad you discovered this workaround.

BuckLocal/ruby_scripts/project_generator.rb
```diff
@@ -89,7 +89,7 @@ def override_xcconfig_files
     # Disable `build_implicit_dependencies` and `parallelize_buildables` for "BuckLocal" xcscheme.
     #
     def update_main_scheme
-      scheme_name = @xcworkspace.split('/').last.split('.').first
+      scheme_name = @xcworkspace.split('/').last.split('-').first
```
shepting, line 92: Do these schemes now have a different format in the new Xcode?

shepting (approved):

BuckLocal/ruby_scripts/project_generator.rb
qyang-nj: I'm curious as well

BuckLocal/remap_debug_source_path.sh
qyang-nj: Curious to know what happened here. Will 1 second always be enough?

xianwen: I don't know if 1 second will always be enough, but it seems to be good enough from a few basic testing I did locally. Without this, there is always an error message when I build the app in Buck Local.

BuckLocal/ruby_scripts/project_generator.rb
xianwen: No, it's the workspace has a different name.
Before it was `ExampleAppBuckLocal.xcworkspace/xcshareddata/xcschemes/ExampleAppBuckLocal`, now, it's `ExampleAppBuckLocal-BUCK.xcworkspace/xcshareddata/xcschemes/ExampleAppBuckLocal`.

We use `-` so that we can remove that `-BUCK` from the workspace name.

PR #153, opened 2021-09-02
Upgrade CryptoSwift so that compile can pass on Xcode 12

xianwen:


qyang-nj (approved):
Thank you!

